PR #256, opened 2026-08-04
refactor: test the Lovelace resource outcome, not the load mechanism

chrreiter:
Closes #207 — by settling it the other way, plus the cleanup around it that does hold up.

## The premise did not survive checking

#207 asks to drop the manual `resources.async_load()` in `_async_lovelace_resources`, on the
reading that `ResourceStorageCollection`'s `async_items` / `async_create_item` /
`async_update_item` / `async_delete_item` each ensure the collection is loaded first.

Three of those four do. **`async_items` does not.** It is not overridden by
`ResourceStorageCollection` at all — it is `ObservableCollection.async_items`:

```python
@callback
def async_items(self) -> list[_ItemT]:
    """Return list of items in collection."""
    return list(self.data.values())
```

Being a *sync* callback, it structurally cannot await a load, so it reports nothing until
something else has loaded the collection. And nothing does at startup: `lovelace.async_setup`
constructs `ResourceStorageCollection` (`loaded = False`) and never loads it — the load is
lazy, driven by the first mutation or the first `lovelace/resources` list from a browser.

Both of our callers read `async_items()` **before** they write.

## So dropping it is a regression, not a cleanup

On a Home Assistant that has served no dashboard yet — i.e. normal startup ordering:

- **Register**: `async_items()` → `[]` → we conclude nothing of ours is registered →
  `async_create_item` (which self-loads, pulling the existing entry into memory) → a
  **second resource for the same module**. The card is fetched twice and the second
  `customElements.define` throws. That is the exact failure `_points_at_card`'s docstring
  exists to prevent.
- **Unregister / remove entry**: `async_items()` → `[]` → nothing matches → our resource is
  orphaned in the user's config after uninstall.

Demonstrated, not argued: with the block removed, both new tests fail, the register one with
`assert [{'res_type': 'module', 'url': '/haventory_static/haventory-card.js?v=0.3.0'}] == []`.

Verified against the declared **2026.6.0** floor (`raw.githubusercontent.com`, tag `2026.6.0`)
and a live **2026.7.4** container. The sync-callback argument is version-independent.

## What this PR does instead

- **Answers the `hasattr` question in the issue**: that guard *is* redundant, and it goes.
  `ResourceStorageCollection.loaded` is `False` and `ResourceYAMLCollection.loaded` is `True`,
  both class attributes; `tests/conftest.py` has no Lovelace stub at all, and every mock in
  the test files defines `.loaded`. Folded into `getattr(resources, "loaded", True)`, which
  keeps the same tolerance for an unknown collection in one check instead of two.
- **Keeps the `resources.loaded = True` write, and explains it.** It is not bookkeeping:
  `StorageCollection.async_load()` does not set the flag — only HA's private
  `_async_ensure_loaded` does — so skipping it makes the collection's next write load again
  and re-add every item to `self.data`, re-firing `CHANGE_ADDED` for each.
- **Retargets the test the issue named.** `test_remove_entry_loads_resources_before_deleting`
  asserted the mechanism (`resources.loaded is True`); it now asserts the outcome — the
  resource is gone from the backing store — under a name that says so.
- **Adds the register-side case, which had no coverage at all.** Every existing mock was
  constructed `loaded=True`, so the duplicate-resource path was never exercised.
- **Both mocks now mirror the real collection**: blind on read until loaded, self-loading on
  mutation. Either test fails if the ensure-load goes.
- **Corrects the note in `docs/card_shipping_plan.md`** that was #207's stated source, so the
  disproved claim stops circulating.

## Testing

Backend gate, all green:

```
PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q     538 passed, 22 skipped
uv run ruff check .                                   All checks passed!
uv run ruff format --check .                          106 files already formatted
uv run mypy                                           Success: no issues found in 14 source files
```

Frontend untouched.

## Note for the maintainer

`docs/open-items.md` item 56 still carries the original framing. Left alone deliberately —
concurrent branches are editing that file — so it needs retiring separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/custom_components/haventory/__init__.py
+++ b/custom_components/haventory/__init__.py
@@ -471,7 +471,14 @@ async def _async_lovelace_resources(hass: HomeAssistant, *, op: str) -> Any:
         )
         return None
 
-    if hasattr(resources, "loaded") and not resources.loaded:
+    # The collection's create/update/delete each load storage on demand, but
+    # `async_items` cannot — it is a sync callback over an in-memory dict, so an
+    # unloaded collection reports no resources at all. Both callers read it
+    # first: register would add a second entry for a card already registered,
+    # unregister would leave ours behind. Setting the flag is part of the
+    # contract rather than bookkeeping — `async_load` leaves it False, and the
+    # collection's next write would then load and re-add every item again.
+    if not getattr(resources, "loaded", True):
         await resources.async_load()
         resources.loaded = True
 
--- a/docs/card_shipping_plan.md
+++ b/docs/card_shipping_plan.md
@@ -136,9 +136,14 @@ option, and nothing in the dogfood plan installs via HACS-from-branch (the dev l
 | R15 | Stray release asset named `haventory-card.js` would bind HACS's download counter | `filename` names the zip, which `zip_release` makes the installed asset | hacs.json diff |
 | R16 | HA's frontend installs its own `CustomElementRegistry` during boot; a definition made before the swap stays in the registry HA replaced, where the dashboard never looks — the card renders as "custom element doesn't exist" | `defineCardElement()` re-asserts the definition on the first observed registry change, within a 15 s window | 5 cold loads per engine, before/after, on a real HACS install of the release asset |
 
-Out-of-scope follow-up (record in `docs/open-items.md` when implementing): the manual
-`resources.async_load()` in `_async_lovelace_resources` is redundant at the 2026.6.0
-floor (collection methods self-ensure-loaded) and can be dropped in a later cleanup.
+Out-of-scope follow-up, since resolved and **rejected**: the manual `resources.async_load()`
+in `_async_lovelace_resources` looked redundant at the 2026.6.0 floor, on the reading that
+the collection methods all self-ensure-loaded. They do not. `ResourceStorageCollection`
+overrides `async_create_item` / `async_update_item` / `async_delete_item` with an
+ensure-loaded, but not `async_items` — that one is `ObservableCollection.async_items`, a
+*sync* callback over an in-memory dict, so it cannot await a load and reports nothing until
+something else has loaded the collection. Lovelace's `async_setup` never does, and both of
+our callers read `async_items` before writing. The load stays; see the comment on it.
 
 ## Touchpoint inventory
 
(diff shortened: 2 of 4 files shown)


chrreiter (commented):
Approach confirmed (verification requested by the owner before merge) — I checked the PR's central claim independently rather than taking the body's word for it.

**Source check at the declared floor.** I pulled the `homeassistant==2026.6.0` wheel from PyPI and read the three files:

- `helpers/collection.py:140-142` — `ObservableCollection.async_items` is a sync callback returning `list(self.data.values())`; `ResourceStorageCollection` does not override it. It structurally cannot load.
- `components/lovelace/resources.py:74-87` — `async_create_item` / `async_update_item` / `async_delete_item` each `await self._async_ensure_loaded()`, exactly as the PR says. So the issue's premise was three-quarters right, and the wrong quarter is the one that matters: both of our callers read before they write.
- `components/lovelace/__init__.py:185` — storage mode constructs `ResourceStorageCollection` (`loaded = False`) and never loads it; the only setup-time `async_load()` is the dashboards collection.

**The kept block is HA core's own idiom, not a workaround.** `components/lovelace/websocket.py:109-111` does the identical `if not resources.loaded: await resources.async_load(); resources.loaded = True` dance before reading `async_items` — including the manual flag write this repo's block was criticised for. And `async_load()` indeed does not set the flag; only the private `_async_ensure_loaded` does (`resources.py:63-67`), which confirms that dropping the `resources.loaded = True` write would make the collection's next mutation re-load and re-add every item.

**Reproduced the demonstration.** On this branch with the guard block removed, exactly the two new tests fail (`test_remove_entry_deletes_from_an_unloaded_collection`, `test_finds_the_existing_entry_in_an_unloaded_collection`); with it restored, all 58 pass in both files. The register-side gap was real — every pre-existing mock was constructed `loaded=True`, so the duplicate-resource path had never been exercised.

The `getattr(resources, "loaded", True)` fold preserves the old tolerance for a collection without the attribute (treated as loaded, load skipped), and the YAML collection's `loaded = True` class attribute means it is skipped correctly too.

Closing #207 as settled-the-other-way, with the outcome-shaped tests and the corrected `card_shipping_plan.md` note, is the right resolution. Good to merge.

---

Issue #280, opened 2026-08-05
Migrate `hass.data[DOMAIN]` runtime wiring to `entry.runtime_data`


chrreiter:
### What & why

HAventory parks its runtime objects — the repository, the `DomainStore`, the rate limiter, the persist lock — in `hass.data[DOMAIN]`, the historical shared-dict convention: untyped (every access is a cast), invisible to HA, created and cleaned up by hand in setup/unload.

Since HA 2024.4 each config entry carries `entry.runtime_data`: assign the runtime object in `async_setup_entry`, read it wherever the entry is available, and HA clears it automatically on unload. With a typed alias (`type HAventoryConfigEntry = ConfigEntry[HAventoryRuntime]`) mypy checks every access. It is the required idiom for new core integrations and the cleaner pattern for custom ones.

The one design touchpoint: `ws_guard`'s loaded-runtime refusal detects "config entry gone" via a `hass.data` lookup today; WS handlers hold no entry reference, so after the switch they resolve the single entry (`hass.config_entries.async_entries(DOMAIN)` — single-instance makes this trivial) and check its state instead.

Scheduled with the V0.7.0 backend glow-up (#230) so the runtime wiring is touched exactly once. Pure internal refactor: no user-visible change, no schema impact.

### Source

`dev/release_review.md` follow-ups + owner direction 2026-08-05.

### Impact

Low (idiom / maintenance)

### Effort

S (≲ half a day)

PR #454, opened 2026-08-14
feat(diagnostics): entry diagnostics, repairs issues and a guarded lossy load

chrreiter:
## Summary

Three HA-idiomatic surfaces the integration was missing, plus the module move the first of
them needs.

- **`diagnostics.py`** — config-entry diagnostics, reachable from the entry's ⋮ menu.
  Aggregates only: `INTEGRATION_VERSION`, the supported and running schema numbers, the
  repository counts, its generation, the index-health verdict, the sorted
  `hass.data[DOMAIN]` **key names**, the entry's options with `card_title` redacted, and
  whether the built card bundle is on disk and how big it is. No item or location bodies at
  any depth.
- **`health.py`** — `_collect_health_issues` and its `_check_*` helpers move out of `ws.py`
  as `collect_health_issues`. Diagnostics and `haventory/health` ask the same question, and
  having the diagnostics platform import the WebSocket layer inverts the dependency.
  `ws_health` is otherwise unchanged.
- **Repairs** — the three setup refusals now put a card in Settings → Repairs beside the
  entry's error state. The two schema refusals are informational (`ERROR`, not fixable);
  the corrupt-row refusal is **fixable** (`WARNING`), and its flow copies the store to
  `.storage/haventory_store_corrupt_backup`, sets a one-boot `allow_lossy_load` option and
  reloads. Setup honours that option, loads the readable remainder, spends the option, and
  clears every issue id whenever a store reads cleanly.

The two bullets #225 also lists were already done and are untouched here:
`single_config_entry: true` is in the manifest (pinned by
`test_manifest_declares_single_config_entry`), and `entry.runtime_data` is split out as
#280 for V0.7.0.

Closes #225

## Decisions taken against the code

The 2026-08-05 implementation notes were written against a much older tree. Where they had
drifted:

1. **The schema issues carry `{error}` + `{storage_key}`, not the two version numbers.**
   The notes asked for "the stored and supported versions as placeholders". Neither
   `SchemaDowngradeError` nor `CorruptSchemaVersionError` carries them structurally — only
   inside the message — and both messages are already written for a user and already name
   whatever disagreed (`schema_downgrade_message`, `_corrupt_schema_version_message`).
   Handing the message over whole keeps one wording for the entry's error state and the
   Repairs card; splitting it would have meant a second wording to keep true. The card
   still says exactly which versions disagree — see the live evidence below.
2. **`store_schema_version` and `supported_schema_version` are equal whenever the entry is
   loaded.** The notes describe them as a pair worth comparing; in the tree
   `DomainStore.schema_version` *is* the version this build supports (it is the constructor
   argument), and a payload on any other version is either migrated or refused before a
   store reaches the bucket. Both are reported anyway: `null` for the store version is the
   signal that setup never got that far, which is the state diagnostics is most useful in.
3. **The backup helper lives in `storage.py`, not `repairs.py`.** `async_backup_store` goes
   straight to Home Assistant's `Store`, deliberately around `DomainStore` — which
   migrates, normalizes and refuses the very payloads worth copying. Keeping it in
   `storage.py` also makes it offline-testable; `repairs.py` cannot be imported offline at
   all, since nothing in the package imports it and HA loads it on demand.
4. **The fix flow finds its entry through `async_entries(DOMAIN)`, not an issue `data`
   payload.** `single_config_entry` means there is exactly one, so the flow carries no id
   that could go stale between the refusal and the button press.
5. **A failed reload aborts the flow instead of completing it.** Home Assistant deletes an
   issue whenever a fix flow ends any way other than `ABORT`, so completing on a reload
   that did not come up would take the card away while the integration is still down.
   Three abort reasons are translated: `reload_failed`, `backup_failed`, `no_config_entry`.
6. **`is_persistent=False` on all three.** Each issue describes the last setup attempt, and
   setup runs every boot; a stored copy would outlive a store the user has since restored.
7. **No new manifest dependency for the repairs platform.** Core integrations that ship a
   `repairs.py` (`anthropic`, `backblaze_b2`, `bthome`) do not declare `repairs`, and
   hassfest on this PR is the check that settles it — as §4 of the session plan says it
   should be.
8. **`haventory.health` joins the per-module strict mypy list.** The code was strict-checked
   inside `ws.py`; moving it out would have quietly dropped it to the baseline.
9. **One wording fix found only by running it.** The confirm step first read "Selecting
   **Submit** …"; Home Assistant renders an empty-schema form's button as **OK**. It now
   says "Confirming copies …", which is true in every language.

## Type of change

## How was this tested?

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` → **1084 passed, 22
  skipped**; `uv run ruff check .` clean; `uv run ruff format --check .` → 166 files already
  formatted; `uv run mypy` → no issues in 24 source files.
  `npx vitest run` → **1843 passed** across 62 files, `npm run build` → 605.51 kB bundle.
  No card change here; the gate runs because the commit gate wants both halves.
- [x] phacc: `scripts/test_integration.sh` → **93 passed** (89 before this PR, plus the 4
  new cases), run in the `ghcr.io/astral-sh/uv:python3.14-bookworm` container this Windows
  host needs.

New tests:

| file | covers |
|---|---|
| `tests/test_health_offline.py` | the healthy verdict, one index drift, and all four count comparisons still firing after the move |
| `tests/test_diagnostics_offline.py` | the payload's shape; **no seeded name, note or custom-field value anywhere in `json.dumps(payload)`**; the redacted title; an unloaded entry; bundle present and absent |
| `tests/test_init_offline.py` (+6) | each refusal raising its own issue with the right severity and fixability; a clean store deleting all three ids; the opt-in loading the remainder and being spent, leaving the other options alone |
| `tests/test_storage_offline.py` (+2) | the backup copying the file verbatim (unreadable rows included) and leaving the original; reporting when there was nothing to copy |
| `tests/integration/test_diagnostics.py` | the platform being reachable over the real `/api/diagnostics/config_entry/…` endpoint, and its output surviving HA's JSON encoder |
| `tests/integration/test_repairs.py` | the downgrade issue landing with the store untouched; the whole fix flow — refuse → offer → copy → opt in → reload → clear; a clean store leaving no issue |

One local-only failure worth naming:
`test_no_unregistered_file_states_an_interpreter_version` fails on this machine because
that test sweeps the working directory and this checkout carries `.claude/settings.local.json`
(ignored by the developer's *global* gitignore, so invisible to the repo and to CI) plus
other sessions' stale `.claude/worktrees/`. Re-run on a clean copy of the same tree inside
the container: **1084 passed, 0 failed**. CI is the final word.

## Live verification (dev Docker HA, real store: 1087 items / 89 locations)

**A — diagnostics download.** The entry's ⋮ menu carries "Download diagnostics" (the German
dev instance shows *Diagnosedaten herunterladen*, between *Eintrag-ID kopieren* and
*Systemoptionen*). `GET /api/diagnostics/config_entry/<id>` → `200`, **3982 bytes** for a
1087-item inventory. Every distinct string in the live store was checked against the
document — 1086 item names, 81 location names, 318 descriptions, 178 custom-field values —
and **none of them appear**. `card_title` reads `**REDACTED**`. `frontend_bundle` reports
`exists: true, size_bytes: 605515`; `health_issues: []`.

**B — a store hand-stamped above `CURRENT_SCHEMA_VERSION`.** HA stopped, `schema_version`
set to 9, HA started. Entry → `setup_error`. Settings → Repairs shows one card, severity
*Fehler*, with no Fix button and only *Ignorieren*:

> **HAventory cannot read the stored data**
>
> HAventory stopped instead of setting up: stored data uses schema version 9, which is
> newer than this build supports (8); HAventory will not downgrade it. Upgrade HAventory to
> a version that understands this data, or restore a backup taken with this version. The
> stored data was left unchanged.
>
> The file is `.storage/haventory_store` in your Home Assistant configuration directory,
> and it has been left exactly as it was. This card clears itself once HAventory starts
> against a store it can read.

The store is byte-identical across the refusal — `md5 a00848a6bd57cb0afa2added42816286`
before and after, still `schema_version: 9`, still 1087 items.

**C — a store with corrupt rows, and the fix.** Two rows keyed by non-uuids injected into
the real 1087-item store. Entry → `setup_error` with the corrupt-store message naming both
ids. Settings → Repairs shows one card, severity *Warnung*, which opens straight into the
fix flow:

> **Load HAventory without the unreadable entries**
>
> HAventory could not read 2 item(s) and 0 location(s), and found 0 location(s) whose parent
> chain loops back on itself. It stopped rather than load the rest — it saves on every
> change, and the first edit would rewrite the file without them.
>
> Confirming copies `.storage/haventory_store` to `.storage/haventory_store_corrupt_backup`
> and then starts HAventory with everything it could read. The unreadable entries stay out,
> and the copy is how you get them back.
>
> To keep them instead, close this dialog and repair those entries in the stored file, or
> restore it from a backup.

Confirming it produced, in order:

- `WARNING … Loading a store this build cannot fully read, as the repair asked`
- `.storage/haventory_store_corrupt_backup` written — **1089 item rows, both corrupt ones
  present** (`not-a-uuid-row`, `second-broken-row`)
- entry → `loaded`; the HAventory sidebar entry came back; Repairs went to *Aktuell stehen
  keine Reparaturen an* and the dialog reported *Das Problem ist behoben!*
- `haventory/stats` → `items_total: 1087` — the readable remainder, the two broken rows out
- `.storage/haventory_store` still holds all **1089** rows: the load destroys nothing, the
  first mutation would
- `core.config_entries` shows `allow_lossy_load` gone and `card_title`, `quick_filters`,
  `sidebar_panel_enabled`, `todo_entity_id` and all nine rate-limit options intact — the
  opt-in was spent, nothing else touched

The dev store was restored from a snapshot afterwards: 1087 items, 89 locations,
`haventory/health` healthy, entry `loaded`, no repairs, backup file removed.

**Screenshots.** Four PNGs were captured for A, B and C (the ⋮ menu, the downgrade card, the
confirm step, the success state). Publishing them to an `assets/` branch — this repo's usual
route, since GitHub's image upload is web-UI only — was refused by the sandbox on this run,
so the card text is transcribed verbatim above instead. The files are on the owner's machine
under the session scratchpad if they should be attached.

## Checklist

- [x] Docs updated where behavior changed — `README.md` Troubleshooting gains where Repairs
  shows up, what the Fix button does, and what the diagnostics download does and does not
  carry.
- [x] WebSocket API unchanged — `ws_health` keeps its result shape, so
  `docs/backend_api_contract.md` and `docs/data_shapes.md` need no edit.
- [x] Essential invariants preserved. No file under `custom_components/haventory/` was
  deleted or renamed — the health helpers move between two modules that both ship — so
  `RETIRED_PATHS` is untouched.

## Follow-ups

- The corrupt-store repair is offered whenever setup refuses, including on a store with *no*
  readable rows at all; confirming it then loads an empty inventory with the copy as the only
  record. Not filed as an issue: it is a strictly better outcome than the refusal it
  replaces, and the copy makes it reversible either way.
- `Repository._debug_get_internal_indexes()` is now called from `health.py` rather than from
  inside `ws.py`. The name still says what it is and no lint rule objects, but a public
  accessor would read better the next time that area is touched.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1076,3 +1076,16 @@ the first public release.
   the file without those entries on the first edit and make the loss permanent. The store is
   left untouched, and the message names the first few affected ids. Fix those entries, or
   restore `haventory_store` from a backup, then reload the integration.
+- **All three of the above also appear in Settings → Repairs**, which is where they are
+  easiest to find: the entry's error state is one line on one screen. The first two are
+  informational — nothing HAventory can do repairs a store it must not touch. The third
+  offers a **Fix** button: it copies `haventory_store` to `haventory_store_corrupt_backup`
+  and then starts HAventory with everything it could read, leaving the unreadable entries
+  out. Take that only if you would rather have the readable remainder than repair the file;
+  the copy is how you get the rest back. The offer applies to one start — a store still
+  broken at the next restart is refused again.
+- **Reporting a bug**: Settings → Devices & services → HAventory → ⋮ → **Download
+  diagnostics** writes a JSON with counts, schema versions, index-health checks, which
+  runtime pieces are loaded and whether the card bundle is deployed. It carries no item or
+  location content — no names, notes or custom-field values — so it is safe to attach to a
+  public issue. If you need the content, use the export instead.
--- a/custom_components/haventory/__init__.py
+++ b/custom_components/haventory/__init__.py
@@ -17,6 +17,7 @@ from homeassistant.config_entries import ConfigEntry
 from homeassistant.core import HomeAssistant
 from homeassistant.exceptions import ConfigEntryError, ConfigEntryNotReady
 from homeassistant.helpers import config_validation as cv
+from homeassistant.helpers import issue_registry as ir
 from homeassistant.loader import async_get_integration
 
 try:
@@ -53,17 +54,23 @@ from . import services as services_mod
 from . import todo_bridge as todo_mod
 from . import ws as ws_mod
 from .const import (
+    CONF_ALLOW_LOSSY_LOAD,
     CONF_CARD_TITLE,
     CONF_QUICK_FILTERS,
     CONF_SIDEBAR_PANEL_ENABLED,
+    CORRUPT_BACKUP_STORAGE_KEY,
     DEFAULT_CARD_TITLE,
     DEFAULT_SIDEBAR_PANEL_ENABLED,
     DOMAIN,
+    ISSUE_CORRUPT_SCHEMA_VERSION,
+    ISSUE_CORRUPT_STORE,
+    ISSUE_SCHEMA_DOWNGRADE,
     PANEL_ELEMENT_NAME,
     PANEL_ICON,
     PANEL_URL_PATH,
     PLATFORMS,
     QUICK_FILTER_KEYS,
+    REPAIR_ISSUE_IDS,
 )
 from .exceptions import CorruptSchemaVersionError, SchemaDowngradeError, StorageError
 from .rate_limit import RateLimitConfig, RateLimiter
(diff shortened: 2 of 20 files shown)


chrreiter:
The four screenshots this PR's body describes, now attached. They live on the `assets-v060-screenshots` branch under `assets/repairs/` — off `main`, since the repository attaches no binaries to it.

Captured against the dev instance (HA 2026.7.4, German frontend) on the build this PR merged.

**Download diagnostics, from the entry's overflow menu**

Config entry overflow menu with Download diagnostics

**The schema-downgrade card — informational, not fixable.** A store stamped 9 against a build that supports 8; the store was left byte-identical.

Repairs card: HAventory cannot read the stored data

**The corrupt-store card's confirm step — fixable.** Two unreadable rows; the step names the copy it takes and where it puts it. This is the step whose button renders **OK** rather than Submit, which is why the wording says "Confirming copies…".

Repairs flow: Load HAventory without the unreadable entries

**After confirming** — the card is gone and the list is empty.

Repairs list empty, success dialog

One correction worth recording next to these, found in the V0.6.0 pre-release audit: the state in the fourth screenshot does not survive a restart. The lossy load never rewrites the store, so a restart with no mutation in between brings both the refusal and the card back — filed as #457, together with two related defects in the same opt-in.

Issue #442, opened 2026-08-13
The toolchain-pin sweep walks the filesystem, so a contributor's untracked files can fail it


chrreiter:
### What & why

`swept_files()` (`tests/test_toolchain_pins.py:169-183`) enumerates candidates by walking the working tree with `iterdir()`, filtered by a hand-maintained denylist of directory names and paths (`SWEPT_SKIP_DIRS`, `SWEPT_SKIP_PATHS`).

The sweep is meant to cover *committed* files — its own docstring says "Every committed text file that could carry a copy of a version" — but a filesystem walk cannot tell committed from untracked. Any untracked file a contributor happens to have in the tree is swept too, and one carrying a version-like string fails the test with a message about a pin having drifted.

The denylist covers the predictable cases (`.venv`, `node_modules`, the cache directories). It cannot cover the unpredictable ones: local scratch notes, a second virtualenv under another name, tool directories like `.claude/`, editor scratch files. Each new one is a test failure that looks like a real pin drift and is fixed by editing the denylist.

### Impact

Contributor-facing only — CI runs on a clean checkout, and it passes there. It cost one false failure during the v0.5.0 review, which is the shape of the problem: the failure names a pin, so the first move is to go looking at pins.

### Notes / acceptance

Listing from the git index — `git ls-files` — makes the set exactly what the docstring says, and lets most of `SWEPT_SKIP_DIRS` go away since git already ignores those. The generated-tree entries in `SWEPT_SKIP_PATHS` that *are* committed still need skipping.

Worth keeping a fallback for a source tree with no `.git` (a release tarball), rather than erroring there.

Done looks like: the sweep covers exactly the tracked files, and an untracked file carrying a version string does not fail the suite.

### Provenance

v0.5.0 release review.

chrreiter:
Shipped by #488 (`753fd7a`, on `main` 2026-08-20): `swept_files()` now reads `git ls-files -z`, the skip lists shrink to what is actually committed (`dev/`, the two lockfiles, the register itself), and `test_a_local_only_file_is_not_swept` writes an untracked probe stating a version and asserts the sweep never reads it — the exact failure this issue describes.

The one acceptance bullet #488 did not take — a fallback for a source tree with no `.git`, such as a release tarball — is not filed further: the release artifact (`haventory.zip`) carries `custom_components/haventory/` only, so no tree without `.git` ever runs this suite. Closing as done.

Issue #229, opened 2026-08-03
Schema collapse to v1 — must ship before any public install


chrreiter:
### What & why

The release that gives every future install a clean schema floor: `CURRENT_SCHEMA_VERSION` → **1** at the final pre-HACS shape, the dev-era migrations deleted. **Must precede any public install** — the safety argument is that nothing 0.x is published beyond the owner's own install, and that argument expires at the HACS release.

**Re-staged 2026-08-11 (owner):** ships in **V0.6.0**, last in the milestone — after the automation work's schema changes, not before them.

The 2026-08-05 staging put this in V0.5.0 on the theory that v0.4.0 had made every known schema change, so v1 would be the final shape. #187 breaks that theory: reminders are specified as "next date + interval (days/weeks/months)", which is new stored state — not derivable from the `due_date` / `inspection_date` already on `Item`. A V0.5.0 collapse would be followed one milestone later by exactly the `v1→v2` it was ordered early to prevent.

The collapse changes no behavior. It restamps a version number and deletes migration steps no future install will ever run; its entire value is not shipping a dev-era migration tree to a user base that has no use for it. That value is the same whenever it lands, so it costs nothing to put it last. Schema changes needed for V0.6.0 work take the version up as usual (v7, v8, …), and the collapse flattens whatever that reaches.

If the shape is still moving when V0.6.0 closes, this moves to V0.7.0 and the adopter's deletion moves to V0.8.0 with it. The one constraint that does not move: **the collapse precedes any public install.**

### The crossing is in-place — no manual user action

The schema exercise (#226 — `migrate_4_to_5`, the item-`status` backfill in v0.3.0) proved the migration machinery on the live production store. Beyond that step, every dev-era migration is a `setdefault` no-op, so a current store is shape-identical (or one deterministic, idempotent backfill away) from the v1 this release defines: there is nothing to convert, only a number to restamp.

So instead of the export → wipe → import crossing originally planned here, `migrations.py` empties to the driver plus **one sunset adopter**: a store stamped inside the known dev range is read, the idempotent backfills are applied defensively, and the payload is saved as v1. The owner's install upgrades like any other release.

Two constraints, and both matter:

- **The adopter is a back door.** While it exists, this build accepts stores it otherwise could not read. It carries exactly one install across and is **deleted one milestone after this ships** (V0.7.0 on the current staging), before any stranger's store can depend on it. It cannot be deleted in the same release that introduces it — it has to exist in a build the owner actually runs.
[...]

PR #469, opened 2026-08-15
fix(storage): spend the lossy-load opt-in on every load and make the repair stick

### LINKED ISSUE #457: The lossy-load opt-in is never spent on a clean load, and the repair it belongs to does not survive a restart
### What happens

Three linked defects in the corrupt-store repair shipped by #454. All three were reproduced on the dev container against a real 1087-item store.

**1. A clean load never spends `allow_lossy_load`.** `_clear_lossy_load_option` is called only inside the `if load_report.has_corruption:` branch (`__init__.py:296-299`). `const.py` says the option is "cleared there... leaving it set would silently accept the next corruption too", but a load that finds nothing wrong never reaches that line, so the option stays on the entry.

Reproduced: wrote `allow_lossy_load: true` into the entry, booted against a clean store — entry loaded, option still present in the options dump afterwards.

**2. A leftover option turns the next corruption into silent data loss.** With the stale option armed, a store that later corrupts loads at WARNING ("Loading a store this build cannot fully read, as the repair asked" — no repair had run), creates **no Repairs card**, writes **no backup**, and drops the unreadable row. The next mutation persists the loss.

Reproduced: clean store + leftover option, then one row corrupted → 1087 items became 1086, `repairs/list_issues` empty, no `haventory_store_corrupt_backup` written. One WARNING line is the entire trail.

**3. The repair does not survive a restart.** The fix flow's lossy load never rewrites the store file, so the corrupt rows stay on disk until some unrelated mutation happens to persist. Restart before that mutation and the refusal and the card come straight back.

Reproduced: ran the fix flow (backup written with both broken rows, entry loaded with 1086 items, card cleared, option spent), then restarted with no mutation in between — refusal again, card again, store still holding the broken row.

### Why it matters

The refusal in #454 exists so a partial inventory is never written over a repairable file. Defect 2 is that guarantee inverted: the one path that loads lossily without asking is also the one that keeps no copy. Defect 3 means a household that repairs, sees it work, and restarts HA that evening is back where it started — with a backup file that now looks like the only thing the repair did.

Defect 1 is what makes 2 reachable: the option is spent correctly on the path it was designed for, so it only lingers when the reload lands on a store that reads fine — which is exactly what happens when the user restores a backup by hand before pressing Fix, or when the flow's reload fails and a later boot succeeds.

### Fix

- Clear the option after **any** successful load, not only a corrupt one — move `_clear_lossy_load_option` out of the `has_corruption` branch in `_async_load_repository`.
- Persist the repository immediately once a lossy load succeeds, so the store on disk matches what was loaded and the repair is durable. The backup taken by the fix flow already preserves the dropped rows, so nothing is lost that was not already set aside.
- Consider taking the backup in the load path rather than only in the repair flow, so a lossy load can never happen without a copy existing.

Done looks like: a clean boot leaves no `allow_lossy_load` on the entry; a corruption arriving with no repair run refuses as designed; and a repaired store still loads after a restart with no mutation in between.

### Notes

Interacts with #229. The collapse touches `async_migrate_if_needed` and the refusal paths this option gates, so the two should not be edited concurrently, and #229's PR should pin that a store inside the adoptable range is adopted rather than routed at the schema cards.

### Provenance

V0.6.0 pre-release audit, 2026-08-15. Reproduced on the dev container (HA 2026.7.4, custom_components at `71e494c`, real 1087-item store); the store was restored afterwards.

chrreiter:
## Summary

Three linked defects in the corrupt-store repair #454 shipped, all reproduced on the dev container during the V0.6.0 pre-release audit.

1. **A clean load never spent `allow_lossy_load`.** `_clear_lossy_load_option` sat inside the `if load_report.has_corruption:` branch, so a load that found nothing wrong left the option on the entry — which is exactly where a hand-restored backup, or a boot after the fix flow's own reload failed, lands.
2. **A leftover option turned the next corruption into silent data loss.** With it armed, a store that later corrupted loaded at WARNING, raised no Repairs card, wrote no backup, and dropped the unreadable row.
3. **The repair did not survive a restart.** The lossy load never rewrote the store, so the corrupt rows stayed on disk and the refusal came straight back on the next start-up.

What this PR does:

- Clears the option after **any** successful load, and only writes the entry back when there is actually something to spend (an ordinary boot no longer touches the entry).
- Adds `_async_settle_lossy_load`: after a lossy load, copy the store aside and persist the readable remainder immediately, so the file matches what was loaded and the repair is durable.
- Takes that copy **in the load path**, not only in the repair flow — so no path can write the readable remainder over the file while the rows it left out exist nowhere. If the copy fails, the store is left exactly as it is and the error says the next start-up will stop again.

The copy uses the store's own key rather than the module default, so the two halves cannot disagree about which file they are working on.

Closes #457

## Type of change

## How was this tested?

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` (1087 passed), `uv run ruff check .`, `uv run ruff format --check .`, `uv run mypy`

Four new offline tests in `tests/test_init_offline.py`:

- a clean load spends a leftover option (defect 1, and the reachability of defect 2);
- an ordinary boot writes nothing back to the entry;
- a lossy load rewrites the store and the dropped rows are in the copy (defect 3);
- a lossy load whose copy fails leaves the store byte-for-byte as it was.

The one deselected test is the known local-only `test_no_unregistered_file_states_an_interpreter_version` (#442) — untracked `.claude/worktrees/` debris in this checkout, not related to this change.

Live check: the `scripts/test_integration.sh` phacc suite runs as its own CI job on this PR; the repairs flow it covers (`tests/integration/test_repairs.py`) exercises the fix flow end to end.

## Checklist

- [x] Docs updated where behavior changed — the option's contract is stated in `const.py`, which now says where it is cleared and what leaving it on would cost.
- [x] No WebSocket API change.

## Follow-ups

None. #229 touches the same refusal paths — per that issue's note, the two should not be edited concurrently; this one lands first.

### DIFF
--- a/custom_components/haventory/__init__.py
+++ b/custom_components/haventory/__init__.py
@@ -79,6 +79,7 @@ from .storage import (
     CURRENT_SCHEMA_VERSION,
     STORAGE_KEY,
     DomainStore,
+    async_backup_store,
     async_persist_immediate,
     cancel_pending_persist,
     read_schema_version,
@@ -155,6 +156,10 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
 
     repository = await _async_load_repository(hass, entry, store)
     hass.data[DOMAIN]["repository"] = repository
+
+    # A load that had to leave rows behind has to reach the file, or the next
+    # restart meets the same rows and refuses all over again.
+    await _async_settle_lossy_load(hass, store, repository)
 
     # Whatever the previous boot left in Settings → Repairs described a store this
     # one just read, so none of it is true any more.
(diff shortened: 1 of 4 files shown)


chrreiter:
Added two phacc integration tests, which are the live evidence for this change — they run against a real Home Assistant core, a real `Store`, real config-entry machinery and the real repairs flow manager:

- `test_the_repair_survives_a_restart_with_no_mutation_in_between` — run the fix flow, assert the store on disk has lost the unreadable row and the copy has kept it, then unload and set up again (a restart, with no mutation in between) and assert the entry loads with no card.
- `test_a_clean_load_spends_an_opt_in_left_on_the_entry` — an opt-in left on the entry is spent by a load that finds nothing wrong.

Both **fail on `main`** and pass on this branch:

```
FAILED tests/integration/test_repairs.py::test_the_repair_survives_a_restart_with_no_mutation_in_between
FAILED tests/integration/test_repairs.py::test_a_clean_load_spends_an_opt_in_left_on_the_entry
2 failed, 3 passed
```

Full local phacc run on this branch: `95 passed in 58.37s`.

The destructive repro against the dev container's real 1087-item store is deliberately **not** repeated: the audit already ran it end to end and restored the instance afterwards, and these tests reproduce all three defects without putting a household's store through it again.

Issue #466, opened 2026-08-15
A stored row with no name loads as an empty-string item instead of counting as unreadable


chrreiter:
### What happens

`load_state` builds an item's name with `name=str(item_data.get("name", ""))` (`repository.py:2438`). A stored row whose `name` key is missing therefore loads as an item named `""` — an item every write path in the codebase would refuse to create.

Reproduced on the dev container: a row with the `name` key deleted loaded cleanly and answered `haventory/item/get` with `"name": ""`. In the same store, a row with `"quantity": "banana"` was correctly counted as corrupt and dropped. So the load report and the model validator disagree about what a valid item is.

Two faults on that one line, not one: `.get("name", "")` returns the default only when the key is **absent**. An explicit `"name": null` returns `None`, and `str(None)` produces the literal string `"None"` — a silently mis-named item, worse than the empty one. `repository.py:2387` has the same pair for locations.

The effect is not confined to display: an empty name produces no search-index tokens, and the to-do bridge would write a line reading ` ×2`.

### Why it matters

Low, and only reachable through a hand-edited or third-party-written store — no path in this codebase can produce it. Four write paths already enforce a non-empty name (`models.py:1009-1010`, `models.py:1122-1126`, `models.py:423-427`, `import_export.py:470-474`); the load path is the only one that does not, and the first mutation after such a load persists the coerced value permanently.

Filing it because the fix is cheap and the inconsistency is the kind that quietly widens: the corrupt-store machinery shipped in #454 exists to make exactly this class of row visible.

### Fix

Add a `validate_item_name` to `models.py` ("non-empty string after trim"), have the existing item and location validators delegate to it, and call it from both `load_state` sites in place of the `str(...)` coercion. It raises `ValidationError`, which both `except` blocks in `load_state` already catch, so the row lands in `dropped_item_ids` and feeds the existing corrupt-store Repairs card with no further wiring.

Enforce **non-empty only, not the length cap** — the load path deliberately grandfathers over-cap stored values elsewhere, and a length check here would refuse stores this codebase itself wrote.

Effort: small — roughly fifteen lines of source and thirty of tests, with no design decision to take: "corrupt" is what all four write paths already say.

Done looks like: a row with a missing, null, non-string or blank name is dropped and reported like any other unreadable row, and the location twin behaves the same.

### Notes

[...]

PR #470, opened 2026-08-15
fix(import): validate reminder and inspection dates, and keep the calendar up on one it cannot read

### LINKED ISSUE #458: Import does not validate reminder or inspection dates, and one bad row takes the whole calendar down
### What happens

`_validate_item_doc` (`import_export.py:465-509`) validates `due_date`, the timestamps, the status and the tags, but never looks at `reminder_date`, `reminder_interval` or `inspection_date` — although `docs/data_shapes.md:565` promises "Dates use `YYYY-MM-DD` and are validated for real calendar dates."

Two ways that lands, both reproduced live:

**A garbage date is accepted and takes the calendar down.** A merge import carrying `"reminder_date": "next week"` passed preview and execute (`items: {update: 1}`) and persisted. From that moment:

- `GET /api/calendars/calendar.haventory` answered **500** for every window.
- Every inventory mutation logged a traceback through the dispatcher — `ValueError: Invalid isoformat string: 'next week'`, raised at `calendar_projection.py:148` and reached from `calendar.py:113`.
- `haventory/reminder/bump` on that item answered `unknown_error`.

`HaventoryCalendar.event` and `async_get_events` have no guard around the projection, so one unreadable row disables the whole entity for every item.

**A typo'd interval is accepted and silently dropped.** `"reminder_interval": {"unit": "month", "count": 3}` (singular) passes import, and the deliberately tolerant `load_reminder_interval` turns it into `None` on load. The document imported clean, the recurrence is gone, and nothing anywhere says so.

The same projection path is reachable from a hand-edited store: a garbage reminder string does not raise inside `load_state`'s try, so it is neither dropped nor counted as corruption.

### Why it matters

An import is the one moment a household hands the integration a document it did not write — a restored backup, a spreadsheet export, another tool's output. It is also the only route back from a newer build. The write paths refuse a bad date; the import path is the one that does not, and the failure it produces is not local to the bad row: the calendar entity is down for everything until the row is found.

Recovery is one `item/update` with `reminder_date: null` (verified — the calendar recovered immediately), but nothing points at the row, and the user-visible symptom is a broken calendar rather than a bad import.

### Fix

Two halves, worth doing together:

1. Validate `reminder_date`, `inspection_date` and `reminder_interval` in `_validate_item_doc`, the way `due_date` already is (`import_export.py:512-526`) — including the `{unit, count}` shape, so a wrong unit is a rejected row rather than a silent loss.
2. Make the projection tolerant of a date it cannot parse: skip that occurrence, log once, keep the entity up. One bad row should cost its own events, not the calendar.

Done looks like: an import naming `"next week"` is refused with the row and field named; a document with `"unit": "month"` is refused rather than quietly stripped; and a store hand-edited to hold a bad date still renders every other item's events.

### Provenance

V0.6.0 pre-release audit, 2026-08-15. Reproduced on the dev container (HA 2026.7.4, `71e494c`) against a real store; the probe row was cleared and the instance restored afterwards.

chrreiter:
## Summary

Two halves of one defect, both from the V0.6.0 pre-release audit.

**Import accepted dates no write path would.** `_validate_item_doc` validated `due_date`, the timestamps, the status and the tags, but never looked at `reminder_date`, `reminder_interval` or `inspection_date` — while `docs/data_shapes.md` promised "Dates use `YYYY-MM-DD` and are validated for real calendar dates". A merge import carrying `"reminder_date": "next week"` passed preview *and* execute and persisted.

**One unreadable row took the whole calendar down.** The projection derives occurrences from stored dates on every read, with no guard, so from that moment `GET /api/calendars/calendar.haventory` answered 500 for every window, every mutation logged a `ValueError: Invalid isoformat string` through the dispatcher, and `reminder/bump` on that item answered `unknown_error`.

Changes:

- `_validate_item_doc` now runs `_validate_inspection_date_doc` and `_validate_reminder_doc` beside the existing due-date check. Errors are attributed per field, so a refusal names the row and the field. The rule binding an interval to an anchor is read from `validate_reminder_rules` rather than restated.
- A misspelled unit (`"month"` for `"months"`) is a **rejected row**, not a silent loss — the tolerant loader would otherwise have stored the item with its recurrence quietly gone.
- `calendar_projection` skips an item whose stored date it cannot parse, keeps the entity up for every other item, and logs once per distinct bad value (the projection runs on every read, so an unguarded warning repeats forever).
- `ws_reminder_bump` names an unreadable stored anchor instead of answering `unknown_error`.
- `docs/data_shapes.md` says what import checks and what an unparsable stored date costs.

The remaining way in is a hand-edited store. Tightening the *load* side — the corruption detector and the model validator disagreeing about what a valid row is — is #466, which is filed for V0.7.0 and covers the same class of defect; this PR deliberately does not widen into it.

Closes #458

## Type of change

## How was this tested?

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` (1096 passed), `uv run ruff check .`, `uv run ruff format --check .`, `uv run mypy`

New offline tests:

- six more rows on the "always enforced rules" parametrize — a garbage reminder date, a date that is not a real calendar day, a garbage inspection date, a misspelled unit, a zero count, an interval with no anchor;
- a valid reminder still previews clean, and a refusal names `items[1].reminder_date` rather than the document;
- the projection: an unreadable reminder costs only that item, the same holds for due and inspection dates, `next_event` survives one, and one bad value is reported once however often it is read;
- `reminder/bump` on a hand-broken anchor answers `validation_error` naming the field.

## Checklist

- [x] Docs updated — `docs/data_shapes.md`, both the import bullet and the validation notes.
- [x] No WebSocket command shape change; the bump refusal is a new message under the existing `validation_error` code.

## Follow-ups

None new. Related: #466 (a stored row missing `name` loads as an empty-string item) is the load-side half of "the corruption detector and the model validator disagree", already filed for V0.7.0.

### DIFF
--- a/custom_components/haventory/calendar_projection.py
+++ b/custom_components/haventory/calendar_projection.py
@@ -14,12 +14,22 @@ occurrences the window covers, and none of them is written anywhere.
 
 from __future__ import annotations
 
+import logging
 from calendar import monthrange
 from collections.abc import Iterable, Iterator
 from dataclasses import dataclass
 from datetime import date, datetime, time, timedelta
 
 from .models import Item, ReminderInterval
+
+LOGGER = logging.getLogger(__name__)
+
+# Which unreadable stored dates have already been reported, as
+# ``(item_id, field, value)``. The projection runs on every calendar read, so an
+# unguarded warning would repeat for as long as the row is there; one line per
+# distinct bad value is what makes the log worth reading. Bounded by the
+# inventory, and only ever holds rows no write path could have produced.
+_REPORTED_UNREADABLE: set[tuple[str, str, str]] = set()
 
 # The dated fields on `Item`, and the word each one contributes to an event's
 # summary. `due_date` only exists while an item is checked out
@@ -127,8 +137,8 @@ def _item_events(item: Item, start: date, end: date, *, limit: int) -> Iterator[
     ):
         if stored is None:
             continue
-        day = date.fromisoformat(stored)
-        if not (start <= day < end):
+        day = _stored_date(item, f"{kind}_date", stored)
+        if day is None or not (start <= day < end):
             continue
         yield _event(item, kind, summary, day, uid=f"{item.id}:{kind}")
 
@@ -145,7 +155,9 @@ def _reminder_events(item: Item, start: date, end: date, *, limit: int) -> Itera
 
     if item.reminder_date is None:
         return
-    anchor = date.fromisoformat(item.reminder_date)
+    anchor = _stored_date(item, "reminder_date", item.reminder_date)
+    if anchor is None:
+        return
     interval = item.reminder_interval
     summary = f"{item.name} reminder"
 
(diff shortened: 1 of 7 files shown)


chrreiter:

```
[PASS] #458: preview refused, naming ['items[1].reminder_date', 'items[2].reminder_interval']
[PASS] #458: calendar REST answered 66 events, not 500
```

A three-row document — one clean row, one carrying `"reminder_date": "next week"`, one carrying `{"unit": "month", "count": 3}` — is refused by `haventory/import/preview`, and the errors name **exactly** those two rows and fields and nothing else. Before this change both rows imported clean. `GET /api/calendars/calendar.haventory` answered a 30-day window with 66 events rather than the 500 the audit reproduced.

PR #471, opened 2026-08-15
fix(events): announce a status reassignment and repaint after a location rename

### LINKED ISSUE #462: Deleting a status with reassign_to rewrites items but fires no event and no repaint
### What happens

`haventory/status/delete` with `reassign_to` rewrites every item carrying the deleted status: `_reassign_status` (`repository.py:358-373`) performs ordinary item edits, so each affected item gets `version + 1` and a fresh `updated_at`.

The handler (`ws.py:2224-2248`) then broadcasts `statuses/deleted`, a payload-less `items/updated` and `stats/counts` — but never calls `notify_mutation`. So for that path:

- no `haventory_item_changed` is fired, for any of the rewritten items;
- the low-stock set is not diffed;
- `SIGNAL_INVENTORY_CHANGED` is not dispatched.

That contradicts two statements in the tree. `docs/backend_api_contract.md:160-167` says `haventory_item_changed` is "fired ... on every path" where an item is mutated, and `events.py:4-7` says "Every mutation path ... calls `notify_mutation`". Both are false here.

### Why it matters

Today the damage is confined to the automation surface: an automation subscribed to `haventory_item_changed` — the intended way to react to inventory edits since #218 — sees nothing while forty items are rewritten underneath it. Anything keeping its own view of item versions falls out of date with no event to tell it.

No sensor count can move on a status reassignment, so nothing visible is wrong on the entity side yet. That is what makes it worth fixing now rather than after it bites: the invariant is stated in two places, one of them the public contract, and the first consumer that depends on it will not be looking for an exception.

### Fix

Call `notify_mutation` for each reassigned item after the persist, with the `updated` action, so the bus event, the low-stock diff and the repaint signal all run the way they do for every other item edit. A bulk reassignment then behaves like the bulk command it is.

If a per-item event for a large reassignment is judged too noisy, the alternative is to say so in the contract and in `events.py` — but silence plus a documented promise of no silence is not an option.

Done looks like: deleting a status with `reassign_to` fires one `haventory_item_changed` per rewritten item, and a test pins the count against the number of items reassigned.

### Provenance

V0.6.0 pre-release audit, 2026-08-15; found by walking every mutation path against the event and signal matrix. Code-verified at `71e494c`, not reproduced live.

### LINKED ISSUE #463: A location rename repaints nothing, so calendar events keep announcing the old path
### What happens

A location rename or re-parent rewrites the denormalized `location_path` on every item underneath it, but fires nothing: `ws_location_update`, `ws_location_move_subtree` and `service_location_update` never dispatch `SIGNAL_INVENTORY_CHANGED`.

Since #452 that signal has a consumer the old reasoning did not anticipate. Every projected calendar event's `description` is the item's `location_path.display_path` (`calendar_projection.py:243`). So after renaming "Garage" to "Workshop", `calendar.haventory`'s current-event attributes keep saying "Garage / Shelf A" until local midnight or until some item happens to be edited.

The README's own automation example templates `trigger.calendar_event.description` to announce where the item is — so the announcement names a location that no longer exists.

### Why it matters

`docs/backend_api_contract.md:179-181` justifies the silence with "no location mutation can move any of the four counts", which is still true. What has changed is that counts are no longer the only thing derived from location data: the calendar renders paths, and it renders them from a cached entity state that nothing invalidates.

A rename is exactly when a household expects everything to catch up — it is a tidy-up action, usually followed by looking at the result.

### Fix

Dispatch `SIGNAL_INVENTORY_CHANGED` from the three handlers that can move an item's path, after the persist. Scope notes from the effort check:

- Only rename and re-parent need it. `location/create` cannot affect an existing item's path, and `location/delete` is refused while the location holds items or children (`repository.py:2057-2062`), so neither can leave a stale path behind.
- The **dispatcher signal** is what should fire, not the bus event. `haventory_item_changed` stays unfired: a derived-path rewrite deliberately does not move an item's `version` or `updated_at`, and the documented action vocabulary has no location word. The rule in the contract stays true; its stated reasoning needs the calendar added to it.
- Gate it on the "did anything actually change" test the handler already computes, so an area-only or no-op save repaints nothing.
- The to-do bridge listens to the two bus events, not this signal, so it runs no extra pass — and correctly, since a to-do line carries no location.
- Cost is four sensor recounts plus one calendar re-derive, on a rare human action. Every WS location handler already calls `get_counts()` once for its `stats/counts` broadcast, so this is the same work a factor of a few, not a new kind of work.

Effort: small — a short helper in `events.py`, three call sites, one doc paragraph, and a handful of tests. The test that would actually have caught it belongs in the integration mode: rename a location, then assert the calendar entity's event description carries the new path.

Done looks like: a rename repaints the calendar at once, `haventory_item_changed` is still not fired for location work, and the contract bullet says why the signal fires anyway.

### Provenance

V0.6.0 pre-release audit, 2026-08-15; code-verified at `71e494c` across `ws.py`, `services.py`, `calendar.py` and `calendar_projection.py`.

chrreiter:
## Summary

Two gaps between the mutation paths and the automation surface #218 shipped. Both were code-verified in the V0.6.0 pre-release audit; both now have a test in the mode that can see them.

**#462 — `status/delete` with `reassign_to` told nobody.** `_reassign_status` performs ordinary item edits: every affected item takes `version + 1` and a fresh `updated_at`. The handler broadcast `statuses/deleted`, a payload-less `items/updated` and `stats/counts`, but never called `notify_mutation` — so no `haventory_item_changed` fired for any of the rewritten items, the low-stock set was not diffed, and the repaint signal was not dispatched. `docs/backend_api_contract.md` and `events.py`'s own docstring both said the event fires on every path; both were false here.

**#463 — a location rename repainted nothing.** A rename or re-parent rewrites the denormalized `location_path` on every item underneath it. Since #452, every projected calendar event's `description` is that path — read from a cached entity state that nothing invalidated, so `calendar.haventory` kept announcing "Garage / Shelf A" after the rename to "Workshop", until local midnight or until some item happened to be edited. The README's own automation example templates exactly that attribute to say where the item is.

Changes:

- `events.py` grows two helpers beside `notify_mutation`, and its docstring stops overclaiming:
  - `notify_bulk_mutation` — one `haventory_item_changed` per item (an automation subscribed to it is watching items, not commands), one low-stock diff and one repaint for the batch (both describe the inventory as a whole; per row they would repaint every sensor once per row).
  - `notify_derived_paths_changed` — the dispatcher signal only. `haventory_item_changed` stays unfired on purpose: a derived-path rewrite deliberately moves neither an item's `version` nor its `updated_at`, and the documented action vocabulary has no location word.
- `Repository.delete_status` returns the ids it moved instead of a count. The WS result keeps its `reassigned: number`; the caller needed the ids to announce them, and re-deriving them at the call site would have been a second copy of the same bookkeeping.
- The repaint is gated on the two edits that actually rewrite a path — name or parent. An area-only reassignment re-anchors the subtree without changing any path, `location/create` cannot affect an existing item's path, and `location/delete` is refused while the location holds anything, so none of them can leave a stale path behind.
- `docs/backend_api_contract.md`: the "Locations fire no bus event" bullet keeps its rule and gains the reason it still dispatches; `status/delete` states the per-item bus behaviour.

Closes #462
Closes #463

## Type of change

## How was this tested?

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` (1093 passed), `uv run ruff check .`, `uv run ruff format --check .`, `uv run mypy`
- [x] phacc: `97 passed` locally (`tests/integration`)

Offline: the reassignment announces every item it rewrote with the version it wrote; a bulk notification repaints once however many items it names; a rename repaints without a bus event; a save that changes no path repaints nothing; `move_subtree` repaints; the service path behaves like the command.

Integration (the mode that can see the entity):

- `test_a_location_rename_reaches_the_calendar_at_once` — rename a location through the service, then read `calendar.haventory`'s `description` attribute. **Verified failing without the fix** (`assert 'Garage' == 'Workshop'`) and passing with it.
- `test_a_status_reassignment_reaches_the_bus_once_per_item` — over a real WebSocket connection, with a real bus listener capturing the events.

## Checklist

- [x] Docs updated — `docs/backend_api_contract.md`, both bullets.
- [x] WS command shapes unchanged (`status/delete` still answers `reassigned: number`).
- [x] Essential invariants preserved — the derived-path rewrite still does not touch `version` or `updated_at`, which is why this fires the signal and not the event.

## Follow-ups

- The to-do bridge listens to the two bus events rather than this signal, so a location rename runs no extra bridge pass — correctly, since a to-do line carries no location.

### DIFF
--- a/custom_components/haventory/events.py
+++ b/custom_components/haventory/events.py
@@ -4,7 +4,13 @@ The WebSocket broadcasts in `ws.py` reach subscribed clients; these reach the
 rest of Home Assistant. Every mutation path — WebSocket handler, `haventory.*`
 service, bulk operation, import — calls `notify_mutation` after its durable
 write, which fires `haventory_item_changed`, diffs the low-stock set to fire
-`haventory_low_stock`, and dispatches the signal the sensors repaint on.
+`haventory_low_stock`, and dispatches the signal the sensors repaint on. A
+command that rewrites many items at once calls `notify_bulk_mutation` instead:
+same events per item, one diff and one repaint for the batch.
+
+A location mutation calls `notify_derived_paths_changed`, which repaints without
+announcing anything on the bus — the items underneath it did not change, only
+the path denormalized onto them.
 
 Bus events bypass the rate limiter: it budgets WebSocket subscription traffic,
 and these are internal to Home Assistant.
@@ -13,6 +19,7 @@ and these are internal to Home Assistant.
 from __future__ import annotations
 
 import logging
+from collections.abc import Sequence
 from typing import Any
 
 from homeassistant.core import HomeAssistant
@@ -78,21 +85,7 @@ def notify_mutation(
             return
 
         if item is not None:
-            _fire(
-                hass,
-                EVENT_ITEM_CHANGED,
-                {
-                    "action": action,
-                    "item_id": item.get("id"),
-                    "name": item.get("name"),
-                    "quantity": item.get("quantity"),
-                    "location_id": item.get("location_id"),
-                    "location_path": (item.get("location_path") or {}).get("display_path"),
-                    "effective_area_id": item.get("effective_area_id"),
-                    "version": item.get("version"),
-                    "ts": iso_utc_now(),
-                },
-            )
+            _fire_item_changed(hass, action, item)
 
         _fire_low_stock_transitions(hass, bucket, item=item)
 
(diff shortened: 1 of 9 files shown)


chrreiter:

```
[PASS] #463: calendar description 'AAA Probe Garage' -> 'AAA Probe Workshop' with no item edit in between
[PASS] #462: status/delete reassigned 1 item and fired 1 haventory_item_changed for it
```

**#463**: a probe item dated today, in a probe location, is `calendar.haventory`'s nearest event; renaming that location through the `haventory.location_update` service moved the entity's cached `description` attribute at once, with no item mutation and no midnight rollover in between — which is exactly what did not happen before.

**#462**: `haventory/status/delete` with `reassign_to` was driven while a second WebSocket connection held a real `subscribe_events` on `haventory_item_changed`. One event arrived for the one reassigned item, with action `updated`. Before this change that connection saw nothing.

PR #473, opened 2026-08-15
feat(services): set, clear and bump reminders from an automation

### LINKED ISSUE #464: Reminders cannot be set or bumped from an automation, and the README documents a call that cannot be made
### What happens

Reminders cannot be set, cleared or bumped from a Home Assistant automation.

`SCHEMA_ITEM_CREATE` and `SCHEMA_ITEM_UPDATE` (`services.py:46-81`) carry `due_date` and `inspection_date` but not `reminder_date` / `reminder_interval`, and there is no reminder service. The WS layer has all of it — `haventory/reminder/set|clear|bump`, plus both fields on `item/create` and `item/update` — but an HA automation cannot send a WebSocket command.

The README (around lines 284-288) tells the user to bump "From an automation or a script that is one WebSocket call" and shows a raw WS frame. That instruction cannot be followed.

### Why it matters

Reminders are the automation-facing feature of the release — "check the smoke detector every three months" is only useful if something can act on it — and #448 made the service surface the automation-facing API. The two shipped in the same milestone without meeting: #453 never extended the schemas #448 had just standardised, and the README documents the gap as if it were the design.

### Fix

Two options; the effort check found the first is smaller than expected because the validation already lives in the model layer.

**(a) Minimal — carry reminders on the existing services.** Add both fields to `SCHEMA_ITEM_CREATE` and `SCHEMA_ITEM_UPDATE` with the same permissive voluptuous shapes the WS commands use (`vol.Any(str, None)` / `vol.Any(dict, None)`), and the matching field entries in `services.yaml`. No new validator is needed: `ws.py` does no shape validation either, it delegates to `validate_reminder_rules`, and the service handlers already catch and re-raise `ValidationError`. No `strings.json` or translation entries are needed — field labels come from `services.yaml`, and the parity test only checks service names. Response data is unchanged: `serialize_item` already emits both fields. Effort: small.

**(b) Full — add `haventory.reminder_bump`** (and optionally `reminder_set` / `reminder_clear`) on top of (a). This is the one the README actually instructs users to call. It needs the new schemas and handlers, rows in the `SERVICES` table, `services.yaml`, and matching `name`/`description` in both `strings.json` and `translations/en.json` (two parity tests force this). One real decision: `ws_reminder_bump` holds the next-occurrence rule in twelve lines; extract it once and call it from both places rather than copying it. Effort: medium.

Either way the README paragraph is replaced with something a household can paste.

Done looks like: an automation can set a reminder and mark it done through `haventory.*` services, and the README shows that automation instead of a WebSocket frame.

### Notes

Also worth fixing while in that README section: the sentence in the calendar/reminders text that says "the **Overdue** sensor is what keeps counting it" describes due dates only — no sensor counts a missed reminder. Filed separately.

No interaction with #229: neither option touches the stored shape, so it is safe on either side of the collapse.

### Provenance

V0.6.0 pre-release audit, 2026-08-15; code-verified at `71e494c` against `services.py`, `services.yaml`, `ws.py`, `models.py` and the README.

### LINKED ISSUE #467: The README calendar section credits the Overdue sensor for reminders it does not count
### What it is

The README's calendar section covers due dates, inspection dates and reminders together, and ends the "a date that has passed" explanation with:

> yesterday's is gone from the calendar, and the **Overdue** sensor is what keeps counting it

That is true for due dates. It is not true for reminders: no sensor counts a missed reminder. `overdue_count` and `inspection_overdue_count` cover `due_date` and `inspection_date` only (`repository.py:1701-1729`), and the sensor names and their own README paragraph are honest about that. A recurring reminder shows only its next occurrence; a one-off whose date has passed shows nothing at all and is counted nowhere.

So a household reading that paragraph would conclude a missed reminder is still visible somewhere, and it is not.

### Fix

Reword so the sentence names what it covers. Either scope it to due dates explicitly, or state the reminder behaviour beside it: a recurring reminder rolls forward on its own and is never "overdue"; a one-off that has passed leaves the calendar and is not counted — bumping or clearing it is the action, and there is no surface that lists it.

Worth doing in the same pass as the reminder-services README paragraph (filed separately), since both sentences are in that part of the file.

### Notes

The underlying behaviour is deliberate and stays as it is. This is a wording fix only — no sensor and no calendar change is being asked for here.

### Provenance

V0.6.0 pre-release audit, 2026-08-15; checked against `repository.py:1701-1729` and the live sensors on the dev container.

chrreiter:
## Summary

Reminders could not be set, cleared or bumped from a Home Assistant automation. `SCHEMA_ITEM_CREATE` / `SCHEMA_ITEM_UPDATE` carried `due_date` and `inspection_date` but not the reminder fields, and there was no reminder service — the whole surface was WebSocket-only, which an HA automation cannot send. The README told the user to bump "From an automation or a script that is one WebSocket call" and showed a raw frame: an instruction that could not be followed.

This takes **option (a) plus the one service the README actually instructs**, per #464's own analysis:

- `reminder_date` and `reminder_interval` join `SCHEMA_ITEM_CREATE` and `SCHEMA_ITEM_UPDATE` with the same permissive `vol.Any(str, None)` / `vol.Any(dict, None)` shapes the WS commands use — the rules live in `validate_reminder_rules`, which names what is wrong far better than a schema mismatch can. Setting and clearing a reminder *are* field writes, so they need no services of their own; `null` on either clears it.
- `haventory.reminder_bump` is the exception, because "I have just done this" is a question about where the series goes next rather than a field write. It answers `{"item": <Item>}` like every other item service, so a script can template `reminder_date` straight out of `response_variable`.
- The twelve-line next-occurrence rule moves out of `ws_reminder_bump` into `calendar_projection.bumped_reminder_date`, which both surfaces call. Two answers to "where does this reminder go next" would be one too many, and the issue asked for exactly this.
- `services.yaml`, `strings.json` and `translations/en.json` all gain the new service and the two new fields (the catalog parity test holds all three to the `SERVICES` table).

**#467 in the same pass**, since it is a sentence in the same README section: the calendar text credited the **Overdue** sensor with keeping a passed date counted. That is true for due and inspection dates and false for reminders — `overdue_count` and `inspection_overdue_count` cover `due_date` and `inspection_date` only. It now says what actually happens: a recurring reminder rolls forward on its own and is never overdue, and a one-off whose date has passed leaves the calendar and is counted nowhere, so bump or clear it while it is in front of you. No behaviour change — the wording was the defect.

Closes #464
Closes #467

## Type of change

- [x] Documentation

## How was this tested?

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` (1112 passed), `uv run ruff check .`, `uv run ruff format --check .`, `uv run mypy`
- [x] phacc: `99 passed` locally

Offline: a reminder set and cleared through `item_update`; a reminder created with the item; the bump service lands where the command lands (series-aligned, not "today plus seven"); it refuses a one-off and an item with no reminder; it reaches the bus as an `updated` item event. The canonical-envelope test now walks the bump too.

Integration: `test_reminders_are_reachable_from_an_automation` drives create → bump → clear through the **real service registry** with `return_response=True` — which is the only mode that catches a handler HA would classify as an executor job and never await. `test_all_services_are_registered` gains `reminder_bump`.

## Checklist

- [x] Docs updated — `README.md` (a pasteable automation, plus the reminder-vs-overdue wording) and `docs/data_shapes.md` (the service-response section).
- [x] No WebSocket command shape change; the WS bump behaves exactly as before, now through the shared helper.

## Follow-ups

None. `reminder_set` / `reminder_clear` services are deliberately not added: they would be `item_update` with fewer fields, and the service picker is a household's menu rather than a mirror of the WebSocket namespace.

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,7 +251,10 @@ running — which for an all-day event means today.
 Nothing is scheduled. The events are worked out whenever something reads the calendar, so
 editing a date changes the calendar immediately and no timer can drift out of step with the
 inventory. A date only exists as an event on its own day: yesterday's is gone from the
-calendar, and the **Overdue** sensor is what keeps counting it.
+calendar. For a **due date** or an **inspection date** the matching sensor keeps counting it
+after that. A **reminder** works differently: a recurring one rolls forward on its own and is
+never overdue, and a one-off whose date has passed simply leaves the calendar — no sensor
+counts it, so bump it or clear it while it is still in front of you.
 
 Notifications are an ordinary calendar automation — the same one you would write for a
 birthday:
@@ -289,17 +292,45 @@ a reminder anchored on the 31st shows 28 February and then 31 March, rather than
 to the 28th forever.
 
 When you have actually changed the filter, **bump** the reminder and the whole series moves
-on one step. From an automation or a script that is one WebSocket call:
+on one step. From an automation or a script that is one service call — say, when the smart
+plug on the boiler reports the service engineer's visit is done:
 
-```json
-{"type": "haventory/reminder/bump", "item_id": "…"}
+```yaml
+automation:
+  - alias: The filter has been changed
+    trigger:
+      platform: state
+      entity_id: input_button.hvac_filter_changed
+    action:
+      - service: haventory.reminder_bump
+        data:
+          item_id: "0f2c…"
+        response_variable: bumped
+      - service: notify.notify
+        data:
+          message: "Next filter change: {{ bumped.item.reminder_date }}"
 ```
+
+Setting and clearing a reminder are ordinary field writes, so they ride the item services:
+`haventory.item_create` and `haventory.item_update` both take `reminder_date` and
+`reminder_interval`, and `null` for either clears it.
+
+```yaml
+      - service: haventory.item_update
+        data:
+          item_id: "0f2c…"
+          reminder_date: "2026-09-01"
+          reminder_interval: { unit: months, count: 3 }
+```
+
+The same three verbs are on the WebSocket API for a client that has one open —
+`haventory/reminder/set`, `/clear` and `/bump`.
 
 Bumping counts from today when the reminder is overdue, so one you forgot for a year lands on
 its next future date rather than on another one already past. "Today" is your Home Assistant
 timezone's day, the same one the calendar rolls over on — so bumping something in the evening
 advances it to the occurrence the calendar is showing you next, wherever you live. A reminder
-you no longer want is cleared from the same editor, or with `haventory/reminder/clear`.
+you no longer want is cleared from the same editor, or by writing `null` over its date.
 
 ### Shopping list
 
--- a/custom_components/haventory/calendar_projection.py
+++ b/custom_components/haventory/calendar_projection.py
@@ -20,6 +20,7 @@ from collections.abc import Iterable, Iterator
 from dataclasses import dataclass
 from datetime import date, datetime, time, timedelta
 
+from .exceptions import ValidationError
 from .models import Item, ReminderInterval
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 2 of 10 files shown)


chrreiter:

```
[PASS] #464: reminder_bump registered, response={'optional': True}, item_update carries reminder fields=True
[PASS] #464: haventory.reminder_bump answered reminder_date=2026-08-21 (future, and whole weeks from the anchor)
```

`get_services` — the only thing that proves a service really declares `SupportsResponse` — shows `haventory.reminder_bump` registered with `response: {optional: true}`, and `haventory.item_update`'s field list carrying `reminder_date` and `reminder_interval`.

Driven end to end through the REST service endpoint with `?return_response`: a reminder anchored 400 days ago with a 7-day interval was bumped and answered `2026-08-21` — in the future, and a whole number of weeks from the anchor, so the series is intact rather than restarted from today. That is the call the README now tells a household to make, made.

PR #475, opened 2026-08-15
fix(diagnostics): keep the household's own vocabulary out of the download

### LINKED ISSUE #465: The diagnostics download carries the household's own status vocabulary
### What happens

The config-entry diagnostics download carries the household's own status vocabulary. `collect_health_issues` feeds `get_counts()` into the payload, and `status_counts` (`repository.py:1683-1697`) is keyed by **every defined status slug** — including the ones a household typed itself.

From the live dev instance's download: `rc_plum`, `night_store`, `sun_faded`, `lent_out` alongside the built-ins. A real household writes `lent_to_alice` or `in_the_caravan`.

`todo_entity_id` is also emitted unredacted, which names the list they mirror to.

### Why it matters

Diagnostics is the file a user is asked to attach to a bug report, on the promise that it carries no inventory. #454 delivered that promise everywhere else — checked against the live download, no item name, description, tag, category, custom field or location name appears at any depth, and `card_title` is redacted. The status slugs are the one thing that slipped through, and they are user-authored free text under a `[a-z0-9_]{1,64}` constraint that does nothing to stop a name being in there.

It is a small leak, but it is the kind that makes the promise itself unreliable — a user who spots their own words in a file they were told was safe has no way to know what else is in it.

### Fix

Report the shape without the vocabulary. Options:

- Emit `status_counts` as a list of counts with the slugs replaced by indices, or as `{count_of_statuses, items_per_status: [...]}` — the diagnostic value is "are items spread across statuses or piled on one", which survives the anonymisation.
- Or redact non-built-in slugs specifically, keeping the ten shipped ones readable and replacing the rest with `custom_1`, `custom_2`.
- Decide explicitly on `todo_entity_id`: it is arguably configuration rather than content, and the module docstring already says it is deliberate — if it stays, the docstring should say why it is different from `card_title`, which is redacted.

Done looks like: a household's own status names do not appear in the diagnostics download, and a test asserts it against a store carrying a custom status.

### Provenance

V0.6.0 pre-release audit, 2026-08-15. Read from the live diagnostics download on the dev container (HA 2026.7.4, `71e494c`, 1087-item store).

chrreiter:
## Summary

The config-entry diagnostics download carried the household's own status vocabulary. `collect_health_issues` feeds `get_counts()` into the payload, and `status_counts` is keyed by **every defined status slug** — including the ones a household typed itself. The live dev download showed `rc_plum`, `night_store`, `sun_faded`, `lent_out` beside the built-ins; a real household writes `lent_to_alice` or `in_the_caravan`, and the `[a-z0-9_]{1,64}` constraint does nothing to stop a name being in there.

#454 delivered the "no inventory content" promise everywhere else — no item name, description, tag, category, custom field or location name appears at any depth. The status slugs were the one thing that slipped through, and it is the kind of leak that makes the promise itself unreliable: a user who spots their own words in a file they were told was safe has no way to know what else is in it.

- `status_counts` now reports the **shape** without the vocabulary: the three slugs this integration ships stay readable, and every other slug becomes `custom_1`, `custom_2`, … carrying its own count. "Are items spread across statuses or piled on one" survives; "which words does this household use" does not. Built-ins stay readable on purpose — a report saying `needs_repair` can be acted on where one saying `custom_2` cannot.
- **`todo_entity_id` is redacted**, which #465 asked to have decided rather than left implicit. It is an entity id rather than prose, but its object id is whatever the household named the list (`todo.alices_shopping`), and the diagnostic question it answers — is the bridge configured at all — is still visible, because redaction keeps the key. The module docstring now states all three redactions and why, instead of claiming the card title is the only authored string in the file.

Closes #465

## Type of change

## How was this tested?

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` (1109 passed), `uv run ruff check .`, `uv run ruff format --check .`, `uv run mypy`
- [x] phacc: `98 passed` locally

The offline fixture now stores a household-authored status and a household-named to-do list, so the existing whole-payload search — the test that stops a future contributor adding the dataset "just for debugging" — covers both. Two new tests pin the shape that survives and the fact that the list is redacted rather than dropped. **All three fail against `main`.**

The integration test downloads through the real HTTP endpoint with a custom status in the store and the to-do option set, and asserts the same.

## Checklist

- [x] Tests added/updated (happy path + edge/error case).
- [x] Docs updated — the module docstring, which is where this contract is stated.
- [x] No WebSocket API change.

## Follow-ups

None.

### DIFF
--- a/custom_components/haventory/diagnostics.py
+++ b/custom_components/haventory/diagnostics.py
@@ -3,14 +3,21 @@
 Home Assistant discovers this module by name and puts "Download diagnostics" on
 the entry's ⋮ menu; the JSON it writes is what a bug report carries.
 
-**Aggregates only — no item or location bodies at any depth.** A diagnostics
-dump answers questions about the *shape* of an install: how much is stored, what
-schema it is on, whether the indexes agree with themselves, whether the card
-bundle is even deployed. Redacting free text well enough to paste into a public
-issue is not something a name, a note or a custom field can be trusted to
-survive, and a user who wants their content already has `haventory/export`. The
-one user-authored string that does appear — the card title, out of the entry's
-options — goes through Home Assistant's redaction helper.
+**Aggregates only — no item or location bodies at any depth, and none of the
+household's own words.** A diagnostics dump answers questions about the *shape*
+of an install: how much is stored, what schema it is on, whether the indexes
+agree with themselves, whether the card bundle is even deployed. Redacting free
+text well enough to paste into a public issue is not something a name, a note or
+a custom field can be trusted to survive, and a user who wants their content
+already has `haventory/export`.
+
+Three strings a household authors could otherwise reach the file, and none does:
+the card title and the chosen shopping list go through Home Assistant's
+redaction helper, and the status slugs — `status_counts` is keyed by *every*
+defined status, and a household writes its own — are replaced with indices
+unless they are the three this integration ships. A user who is told a file is
+safe to attach and finds their own words in it has no way to know what else is
+in there.
 
 Home Assistant can call this on an entry whose setup failed, which is exactly
 when it is worth having, so every block reports what it found rather than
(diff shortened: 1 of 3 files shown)


chrreiter:

```
[PASS] #465: status_counts={"custom_1": 0, "custom_2": 2, "missing": 0, "needs_repair": 36,
                            "custom_3": 2, "ok": 1031, "custom_4": 7, "custom_5": 2,
                            "custom_6": 2, "custom_7": 4, "custom_8": 2}
[INFO] #465: options block: {"card_title": "**REDACTED**", …, "todo_entity_id": ""}
```

Downloaded through the real `/api/diagnostics/config_entry/{entry_id}` endpoint against the dev store, which carries eight household-authored statuses. None of `rc_plum`, `night_store`, `sun_faded` or `lent_out` appears anywhere in the file — the audit read all four out of this same download. The three built-ins stay readable and every count survives, so "36 items need repair, and the rest are spread thinly across eight other statuses" is still answerable from the file.

`todo_entity_id` is unset on this instance and shows as `""`: Home Assistant's redaction helper leaves a falsy value alone, which is the right answer — there is nothing to hide and "not configured" is still visible.

PR #476, opened 2026-08-15
feat(reminders): store the series anchor so a bump keeps its day of month

### LINKED ISSUE #460: A bump rewrites the anchor, so a month-end reminder loses its day permanently
### What happens

The projection's month arithmetic is anchor-measured and correct: a series anchored on the 31st returns to the 31st in every month that has one. Verified live — a `{months, 1}` reminder anchored 2026-08-31 projects 08-31, 09-30, 10-31, 11-30, 12-31.

`haventory/reminder/bump` then destroys that, because it **rewrites the anchor** to the occurrence it computed. Bumping the series above once, in a 30-day month, re-anchored it at 2026-09-30 — and the series became 09-30, 10-30, 11-30, 12-30. The 31st is gone, and no later bump brings it back.

February is the worst case: a 1/31 series bumped in February anchors at 2/28 and is a 28th series from then on.

### Why it matters

"Month steps measured from the anchor so a series on the 31st doesn't drift to the 28th forever" is the guarantee #453 set out to provide, and the projection honours it. But the guarantee only holds for a series nobody bumps, and bump is the feature's main verb — it is what "I have just done this" means. The household that uses reminders as intended is exactly the one that loses the behaviour.

The drift is silent and permanent: the calendar keeps showing a plausible series, one day earlier every time it crosses a short month.

### Fix

Keep the anchor's day-of-month across a bump. The stored `reminder_date` is both "when it next comes round" and "what the series is measured from", and bump currently collapses those into one value.

Two shapes, worth a decision rather than a default:

- Store the series' origin separately from the next occurrence, so bump advances the occurrence and leaves the origin alone. Costs a stored field and a schema bump — awkward directly before #229.
- Have bump re-derive the next occurrence from the **stored** anchor by counting whole intervals forward, rather than writing back the clamped date. A series anchored 1/31 bumped in February then advances to 3/31, not 2/28, because the anchor never moves off the 31st. This needs no new field: the anchor stays the origin, and "next occurrence" is always derived.

The second keeps the derived-on-read principle the calendar already follows and avoids touching the stored shape before the collapse.

Done looks like: a `{months, 1}` series anchored on the 31st still lands on the 31st after any number of bumps, and a test pins a bump taken in a 30-day month.

### Notes

Related to the bump's other date defect — it counts from UTC today while the calendar runs on local days — filed separately. Both live in `ws_reminder_bump` (`ws.py:1395-1407`) and should be fixed in one pass.

### Provenance

V0.6.0 pre-release audit, 2026-08-15. Reproduced on the dev container (HA 2026.7.4, `71e494c`) by reading `calendar.haventory` over the REST API before and after a bump; probe items removed afterwards.

chrreiter:
## Summary

Option A from the analysis on #460, chosen by the owner.

The projection's month arithmetic is anchor-measured and correct — a series on the 31st returns to the 31st in every month that has one. `reminder/bump` destroyed that, because one stored date was carrying two meanings: the next occurrence *and* the series origin. Writing the occurrence back re-anchored the series on it, so a single pass through a 30-day month made it a 30th series and February made it a 28th series, permanently. Bump is the feature's main verb, so the guarantee held only for a reminder nobody used.

**The two meanings are now two fields.**

- `reminder_date` — the next occurrence nobody has marked done. What the calendar shows first, what a household picks, what a bump advances.
- `reminder_anchor` — what the month steps are counted from. Equal to `reminder_date` until the first bump; `null` exactly when `reminder_date` is.

Every path that writes a date anchors the series on it, because picking a date is saying where the series starts. `Repository.bump_reminder` is the one writer that moves the date and keeps the anchor — and it is a repository verb now, beside `check_in` / `check_out` / `set_quantity`, so the WebSocket command and `haventory.reminder_bump` cannot answer differently. The projection walks from the anchor and starts at the date, so occurrences already marked done never come back, even in a window that covers them.

**No occurrence is skipped.** This is the part that ruled out the issue's own option 2: a 31st series bumped in February lands on **28 February**, and the one after it is **31 March**. Advancing straight to 31 March would have kept the day at the price of never reminding the household in February.

**Schema 8 → 9.** `migrate_8_to_9` gives every reminder an anchor equal to its own date — what a series that has never been bumped under the new rule has, and the only honest answer for a v8 series whose drift nothing recorded. Written rather than left absent, for the reason `migrate_7_to_8` gives: a v8 build would otherwise parse these items and rewrite them with every anchor dropped. A pre-v9 store with no anchor reads the same way through `load_reminder_anchor`, so the migration and the load path agree; an anchor stored beyond its own date reads as the date rather than describing a series with nothing to lead to.

**Derived on write, not client-written.** `reminder_anchor` is absent from `ItemCreate` and `ItemUpdate`, and read-only in the card's `Item` type. Export documents *do* carry it, so a backup restores a series mid-flight; on import an anchor later than its own date, or one with no date to lead to, is refused, and an absent one compares as unchanged against a stored series that has never been bumped — so every export written before v9 still round-trips clean.

Closes #460

## Type of change

- [x] Breaking change — the stored schema moves to 9. In-place upgrade only; a v9 store is refused by a v8 build, which is the point.

## How was this tested?

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` (1135 passed), `uv run ruff check .`, `uv run ruff format --check .`, `uv run mypy`
- [x] Frontend: `npx eslint .`, `npm run typecheck`, `npx vitest run` (1843 passed), `npm run build`
- [x] phacc: `102 passed` locally

Offline, the behaviour the issue asked for:

- a 31st series bumped in a 30-day month lands on the 30th **and the next bump lands on the 31st** — the anchor visibly does not move;
- the February worst case, both steps: 1/31 → 2/28 → 3/31;
- setting a date re-anchors on it; clearing takes the anchor with it; a stale `expected_version` still answers `conflict`;
- the projection: a bumped series keeps the anchor's day, occurrences already done stay off the calendar, and an item with no anchor projects from its own date (every pre-v9 store);
- `migrate_8_to_9` anchors on the date, keeps an anchor a bump has moved away from, is idempotent, and leaves a corrupt row alone;
- export/import: a bumped series survives a round trip; a document written before the field existed imports as **unchanged**; three bad anchors are refused by path.

Integration, against a real `Store` and a real WebSocket connection:

- `test_a_v8_store_gains_an_anchor_for_every_reminder_it_holds` — the upgrade the previous release leaves behind.
- `test_a_bumped_month_end_series_keeps_its_day_across_a_reload` — set 31 August, bump, assert the wire and the file, then unload and set the entry up again from what was written, and read `calendar.get_events` for October: `["2026-10-31"]`. **Verified failing** with the anchor-preserving line removed: `assert ('2026-09-30', '2026-09-30') == ('2026-09-30', '2026-08-31')`, which is the defect exactly as reported.

## Checklist

- [x] Docs updated — `docs/data_shapes.md` (the three-field shape, the derived-on-write rule, both migrations), `docs/backend_api_contract.md` (what bump moves and what it does not), `README.md` (the month-end promise now says "however often you bump it").
- [x] `ws.py`, `docs/backend_api_contract.md` and `docs/data_shapes.md` in sync.
- [x] Essential invariants preserved. No file deleted or renamed under `custom_components/haventory/`, so `RETIRED_PATHS` is untouched.

## Follow-ups — this changes S5's brief

`dev/V0_6_0_implementation.md` §5.5 is updated in this PR, because #229's collapse now has one more version to fold in:

- `CURRENT_SCHEMA_VERSION` is **9**, so the closed adoptable set is **2–9**.
- The adopter folds in `migrate_8_to_9` alongside the other backfills — the cheapest of them, since a v8 reminder's anchor is its own date.
- The pre-collapse export the release notes tell the owner to take is stamped **9**, and the schema-downgrade Repairs card's text becomes "(9)" → "(1)".

I will add a short comment to #229 pointing at this, since it changes what that issue asked for.

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,10 +289,12 @@ leave the repeat empty and it is a single date.
 
 The calendar then shows the next occurrence and the ones after it, as far ahead as whatever
 view you are looking at. Nothing is scheduled and no series is written down: the item stores
-one date and one interval however long it runs, and the occurrences are worked out when
-something reads them. A repeat measured in months keeps the day of the month it started on —
-a reminder anchored on the 31st shows 28 February and then 31 March, rather than sliding down
-to the 28th forever.
+one date, one anchor and one interval however long it runs, and the occurrences are worked
+out when something reads them. A repeat measured in months keeps the day of the month it
+started on — a reminder anchored on the 31st shows 28 February and then 31 March, rather than
+sliding down to the 28th forever — and it keeps it **however often you bump it**: bumping a
+31st reminder in a 30-day month moves it to the 30th, and the one after that is the 31st
+again.
 
 When you have actually changed the filter, **bump** the reminder and the whole series moves
 on one step. From an automation or a script that is one service call — say, when the smart
--- a/cards/haventory-card/src/store/types.ts
+++ b/cards/haventory-card/src/store/types.ts
@@ -113,10 +113,16 @@ export interface Item {
   inspection_date: string | null;
   /**
    * Optional because a backend older than schema v8 does not send them; absent
-   * reads as no reminder. The anchor alone is a one-off; with an interval it is
-   * the start of a series the calendar expands on read.
+   * reads as no reminder. `reminder_date` alone is a one-off; with an interval
+   * it is the next occurrence of a series the calendar expands on read.
+   *
+   * `reminder_anchor` is what that series is measured from, and only the
+   * backend writes it — the editor sends a date, and the backend anchors the
+   * series on it. Read-only here, and absent from `ItemCreate` / `ItemUpdate`
+   * for that reason.
    */
   reminder_date?: string | null;
+  reminder_anchor?: string | null;
   reminder_interval?: ReminderInterval | null;
   location_id: string | null;
   tags: string[];
--- a/custom_components/haventory/calendar_projection.py
+++ b/custom_components/haventory/calendar_projection.py
@@ -20,7 +20,6 @@ from collections.abc import Iterable, Iterator
 from dataclasses import dataclass
 from datetime import date, datetime, time, timedelta
 
-from .exceptions import ValidationError
 from .models import Item, ReminderInterval
 
 LOGGER = logging.getLogger(__name__)
@@ -124,40 +123,6 @@ def next_occurrence_after(
     return _occurrence(anchor, interval, _first_step_after(anchor, interval, after))
 
 
-def bumped_reminder_date(item: Item, *, today: date) -> str:
-    """The stored anchor an item's reminder moves to when it is marked done.
-
-    The whole rule, in one place, because two surfaces ask for it — the
-    WebSocket command and the `haventory.reminder_bump` service — and two
-    answers to "where does this reminder go next" would be one answer too many.
-
-    Counted from the later of the anchor and `today`, so a reminder bumped on
-    the day it came round advances by exactly one interval, and one nobody
-    bumped for a year lands on its next *future* occurrence instead of another
-    date already past. `today` is the caller's to supply: it is the household's
-    day, and this module does not know what timezone they live in.
-    """
-
-    if item.reminder_date is None:
-        raise ValidationError("item has no reminder to bump")
-    try:
-        anchor = date.fromisoformat(item.reminder_date)
-    except ValueError as exc:
-        # Only a hand-edited store can hold one, and naming the field beats the
-        # `unknown_error` a raw parse failure answers with.
-        raise ValidationError(
-            f"stored reminder_date {item.reminder_date!r} is not a date this build can read; "
-            "set the reminder again to replace it"
-        ) from exc
-
-    following = next_occurrence_after(anchor, item.reminder_interval, max(anchor, today))
-    if following is None:
-        raise ValidationError(
-            "a reminder with no interval has no next occurrence; clear it instead"
-        )
-    return following.isoformat()
-
-
 def _iter_events(
     items: Iterable[Item], start: date, end: date, *, limit: int = MAX_REMINDER_OCCURRENCES
 ) -> Iterator[ProjectedEvent]:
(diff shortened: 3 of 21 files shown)


chrreiter:
## Live verification — dev container, this branch, against the real 1087-item store

Deployed to the dev Home Assistant (2026.7.4). The store was on **v8** with no reminders, so two probe items were given reminders **on the v8 build first**, to make the migration carry something real. Both removed afterwards; final state `items 1087, locations 89, healthy True, schema 9`; `log_sweep.py --since 15m` → `BLOCKING: 0` (`verdict: PASS`).

**The v8 → v9 upgrade, on 1089 real rows:**

```
BEFORE  schema_version 8 items 1089
  ZZZ probe month-end | reminder_date 2026-08-31 | interval {'unit': 'months', 'count': 1} | has anchor key: False
  ZZZ probe weekly    | reminder_date 2026-09-04 | interval {'unit': 'weeks', 'count': 2}  | has anchor key: False

AFTER   schema_version 9 items 1089
items missing the anchor key: 0
reminders whose anchor is not their own date: []
  ZZZ probe month-end | date 2026-08-31 | anchor 2026-08-31 | interval {'unit': 'months', 'count': 1}
  ZZZ probe weekly    | date 2026-09-04 | anchor 2026-09-04 | interval {'unit': 'weeks', 'count': 2}
```

Lossless, every row backfilled, no item lost, and every existing reminder anchored on its own date exactly as `migrate_8_to_9` says.

**The defect itself, gone, on the instance:**

```
after bump 1:  date 2026-09-30 | anchor 2026-08-31
after bump 2:  date 2026-10-31 | anchor 2026-08-31
```

The first bump lands on 30 September — September has no 31st, and skipping to October would have cost the household that reminder — and the anchor does not follow it. The second bump is back on the **31st**, which is precisely what the audit recorded as lost forever.

And the surface a household actually reads, `GET /api/calendars/calendar.haventory` after two bumps:

```
  2026-10-31 ZZZ probe month-end reminder
  2026-11-30 ZZZ probe month-end reminder
  2026-12-31 ZZZ probe month-end reminder
  2027-01-31 ZZZ probe month-end reminder
```

A 31st series still. On `main` this reads 10/30, 11/30, 12/30, 1/30 — the drift the issue reported.

PR #495, opened 2026-08-20
test(brand): bind the card's mark and the social preview to one geometry

### LINKED ISSUE #285: Brand mark geometry is hand-maintained in two winding conventions with nothing binding them
### What & why

The mark exists as two independent, hand-maintained spellings of the same geometry:

- `cards/haventory-card/src/ui/brand-icon.ts` — `HOUSE` + `CRATES` + `HANDLES` joined into `HAVENTORY_MARK_PATH` (`:55`). The crates are wound **counter-clockwise** on purpose: `ha-svg-icon` renders one `<path d>` and sets no `fill-rule`, so the `nonzero` default only treats a subpath as a hole when it is wound against the shape enclosing it. The file's own comment (`:48-54`) says exactly this.
- `docs/assets/social-preview.html:145` — one `d` attribute, every subpath wound **clockwise**, with `fill-rule="evenodd"` (`:144`), under which winding is irrelevant.

Both render the same mark, and neither is wrong — they are the same shape expressed for two different fill rules. The house subpath is byte-identical between the two; the crates and handles diverge in winding.

The cost is maintenance, not correctness: a geometry change must be made twice, in two conventions that look contradictory side by side, with no test asserting the results still match. #196 adds a third copy — the raster assets for the `home-assistant/brands` PR.

### Approach

Options, cheapest first:

1. **Cross-reference comments** in both files stating that the other exists and why the winding differs. Costs nothing, catches nobody.
2. **Derive the preview from the source** — generate `social-preview.html`'s path from the TS constants at build time, making `brand-icon.ts` the single authority.
3. **A test that normalizes both paths to one winding and asserts equality**, in the spirit of `tests/test_min_ha_version.py` (which exists precisely to stop a value copied into several files from drifting).

Recommend (2) if the preview is ever regenerated, (1) plus (3) if it is not. What is worth avoiding is a third hand-maintained copy landing in #196 without deciding this first.

### Source

Implementation-notes fan-out 2026-08-05 — surfaced while grounding #196.

### Impact

Low

### Effort

S

### Notes / acceptance

- Not a defect: no user sees a wrong mark today. This is about the third copy #196 is about to add.
- `tests/test_frontend_registration.py` already pins `PANEL_ICON` and the bundle's exported identifier to the same string across the language boundary — the same technique applies here.

chrreiter:
## Summary

The mark exists as two hand-maintained spellings of one outline, written for two
different fill rules, with nothing binding them — and #196 was about to add a third as
hand-exported brand PNGs. This closes that off from both ends.

**The binding (#285, options 1 + 3).** `tests/test_brand_assets.py` parses both `d`
strings, normalises each subpath to one winding *and* one start vertex, and asserts the
two describe the same outline. Both files now cross-reference the other and say why the
winding differs.

Normalising winding away is what makes the comparison possible, and it is also what
makes it blind to the one thing each file's fill rule depends on. So two more tests pin
that directly: the card's crates must run *against* its house (`nonzero` cuts a hole
only for a subpath wound against the one enclosing it) and the preview's seven subpaths
must all run alike beside its `fill-rule="evenodd"`. Rewinding a crate to "tidy" the
card to one convention leaves the outline identical and turns the crates into solid
blocks — the shape test cannot see it, these can.

**The brands prep (#196, artwork half).** `scripts/render_brand_assets.py` reads the
card's `HOUSE`/`CRATES`/`HANDLES` constants and writes `docs/assets/brand/`: `icon.svg`
at the mark's own viewBox, plus `icon.png` (256×256) and `icon@2x.png` (512×512). So the
brands artwork is a *rendering* of the card's geometry, not a third copy of it, and
`test_the_brand_svg_is_the_mark_the_card_publishes` fails the moment it goes stale.

The external `home-assistant/brands` PR is **not** opened here — it goes to another
project under the owner's name. Its body and the exact `gh` sequence are below.

Closes #285
Refs #196

## Type of change

Nothing that ships changes. The only edit to card source is a comment, and the built
bundle is byte-identical with and without it (`md5 a1542eca4fdda842c1917c7025fe012b`
both ways — see *How was this tested?*).

#196's 2026-08-05 implementation notes were written before this work was scoped into
V0.7.0 §6.1, which asks for the rasters themselves, rendered from the TypeScript. Where
the two disagree the code and §6.1 win; the issue is left unedited.

1. **`docs/assets/brand/` rather than `docs/assets/brand-icon.svg`.** The notes propose
   one committed SVG as "the export source", with the PNGs living only in the brands PR.
   §6.1 asks for the rasters here, rendered from the TypeScript. Committing them means
   the brands PR is a file copy the owner can inspect before filing, and it means CI
   checks the artwork rather than checking a source nothing renders. The SVG is still
   there, as a generated intermediate rather than a hand-maintained file.
2. **The renderer reads `brand-icon.ts`, not an SVG.** The notes' pin is
   "`brand-icon.svg`'s viewBox equals `HAVENTORY_MARK_VIEW_BOX`". That leaves the `d`
   itself unpinned, which is the copy that actually drifts. Here the whole `d` is
   compared, and the SVG is written from the TypeScript in the first place.
3. **`icon.png` + `icon@2x.png` only — no `logo*.png`, no `dark_*`.** Verified against
   the brands README today: `logo` is for a separate wordmark, "aspect ratio should
   respect the logo of the brand". HAventory's wordmark exists only as live text in a
   font stack in `social-preview.html`; rendering it to a PNG would create exactly the
   hand-maintained asset this PR exists to avoid. `dark_*` is served only when present
   and falls back to the plain one otherwise — the mark is a single mid-tone blue that
   reads on both grounds, so a second variant would be two files to keep in step for no
   gain.
4. **Square *and* trimmed, resolved by fitting the longer axis.** Brands asks for 1:1
   and for "the minimum amount of empty space". The mark is 420×402 in user units, so
   the two are only satisfiable together: it fills the canvas left-to-right and is
   centred vertically, leaving 5px (256) / 11px (512) of transparency top and bottom.
5. **Not interlaced.** The README prefers it; a hand-rolled Adam7 writer costs more code
   and compresses worse, and nothing rejects a non-interlaced PNG. Files are 3.7 kB and
   7.8 kB.
6. **`ignore: brands` in `.github/workflows/ci.yml` is untouched.** The notes pair
   dropping it with a test written failing. That belongs in the PR that lands *after*
   the brands entry exists, and #196 keeps it.
7. **No re-render-and-compare-pixels test.** The chain is closed without one: the SVG's
   `d` is pinned to the TypeScript, and the renderer writes the SVG and both PNGs in a
   single run, so PNGs stale against the SVG would need someone to run half a script.
   A pixel comparison would additionally tie CI to cross-platform `libm` behaviour.
   `test_the_brand_directory_holds_only_what_the_renderer_writes` covers the other
   direction — a hand-made file appearing in that directory.

## How was this tested?

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` → **1164 passed, 22
      skipped**; `uv run ruff check .` → all checks passed; `uv run ruff format --check .`
      → 168 files already formatted; `uv run mypy` → no issues in 24 source files.
- [x] Frontend (`cards/haventory-card`): `npm audit --audit-level=high` → 0
      vulnerabilities; `npx eslint .` → clean; `npm run typecheck` → clean;
      `npx vitest run` → **1860 passed (63 files)**; `npm run build` → 610.08 kB.
- phacc not run: this PR touches no `custom_components/` source and no
  `tests/integration/`. §6.1 requires it for PRs 2–4.

**The new tests are not vacuous** — each was watched to fail against a deliberate
mutation, then the mutation reverted:

| mutation | what fails | what still passes |
|---|---|---|
| second crate's `M148,294` → `M148,296` in the card | `test_a_shifted_corner_is_not_one_shape` (this one is committed, not a manual run) | — |
| third crate moved 2px in `social-preview.html` | `test_the_card_mark_and_the_social_preview_describe_one_shape` | the winding tests |
| first crate rewound to match the house — same outline, opposite direction | `test_the_cards_crates_are_wound_against_its_house`, `test_the_brand_svg_is_the_mark_the_card_publishes` | **the shape test**, which is the point: it proves the normalisation works *and* that it cannot police winding |

**The rasters were read back and checked**, not just eyeballed: `icon.png` 256×256 and
`icon@2x.png` 512×512, bit depth 8, colour type 6 (RGBA), non-interlaced; exactly one
RGB triple present, `(92, 161, 242)` = `#5CA1F2`; alpha bounding box `x=[0,255]`
/ `x=[0,511]` — the artwork touches both vertical canvas edges, i.e. trimmed — and
`y=[5,250]` / `y=[11,500]`.

**The renderer is deterministic**: a second run over the same source produced
byte-identical files (md5 unchanged for all three).

**Nothing shipped changed.** Built the bundle with the new comment, stashed the comment,
rebuilt, restored, rebuilt: `custom_components/haventory/www/haventory-card.js` hashed
`a1542eca4fdda842c1917c7025fe012b` all three times. The comment is stripped by the
minifier, so no live check on the dev HA can distinguish this branch from `main` — which
is why §6.1's live checks all belong to PRs 2–4.

## Checklist

- [x] Tests added/updated (happy path + edge/error cases — see the mutation table).
- [x] Docs updated where behavior changed (`README.md`: how to regenerate the brand
      artwork, beside the existing social-preview line).
- [x] WebSocket API unchanged.
- [x] Essential invariants untouched.

None — the card renders identically (byte-identical bundle, above). The generated
artwork, at 512:

icon@2x

---

## For the owner: the `home-assistant/brands` PR

**Do not** let a session file this — it goes to another project's repository under your
name. The artwork is in this PR; everything below is ready to paste.

### The `gh` sequence

Run after this PR merges, so `main` carries the artwork.

```bash
# 1. Fork and clone (skip the fork if you already have one).
gh repo fork home-assistant/brands --clone --remote=false
cd brands
git checkout -b add-haventory

# 2. Copy the two rasters in. The directory name must equal the manifest domain.
mkdir -p custom_integrations/haventory
curl -sL -o custom_integrations/haventory/icon.png \
  https://raw.githubusercontent.com/chrreiter/HAventory/main/docs/assets/brand/icon.png
curl -sL -o "custom_integrations/haventory/icon@2x.png" \
  "https://raw.githubusercontent.com/chrreiter/HAventory/main/docs/assets/brand/icon@2x.png"

# 3. Sanity-check what you are about to file.
file custom_integrations/haventory/*.png   # want: PNG, 256 x 256 and 512 x 512, RGBA

# 4. Commit and open the PR.
git add custom_integrations/haventory
git commit -m "Add HAventory custom integration icons"
git push -u origin add-haventory
gh pr create --repo home-assistant/brands \
  --title "Add HAventory custom integration icons" \
  --body-file brands-pr-body.md
```

### The PR body (`brands-pr-body.md`)

```markdown
Adds icons for the HAventory custom integration.

- Integration: https://github.com/chrreiter/HAventory
- Domain: `haventory` (matches `manifest.json` and the directory added here)
- Distribution: HACS custom repository today, `hacs/default` submission to follow

`icon.png` is 256x256 and `icon@2x.png` is 512x512, both PNG with transparency and
trimmed to the artwork. No `logo` variant: the integration has no separate wordmark. No
`dark_` variants: the mark is a single mid-tone blue that reads on light and dark
backgrounds alike.

The artwork is generated from the integration's own icon geometry rather than exported
by hand, and a test in that repository fails if the two ever diverge.
```

### After it merges

Drop `ignore: brands` from the `hacs/action` step in `.github/workflows/ci.yml` — that
is #196's remaining half, together with the store submission and the README switch.


chrreiter:
**Superseded: the artwork does not need an external PR to appear.**

The *"For the owner: the `home-assistant/brands` PR"* section of the body above is no
longer the way this artwork reaches Home Assistant, and the `docs/assets/brand/` path it
names no longer exists.

Home Assistant 2026.3 added a brands proxy: a custom integration can carry its own
`brand/` directory and Home Assistant serves those files at
`/api/brands/integration/<domain>/<file>`, taking them **in preference to the brands CDN**
— [announcement](https://developers.home-assistant.io/blog/2026/02/24/brands-proxy-api/).
That release is three below this project's minimum, so every supported install can use it.

#502 moves the artwork to `custom_components/haventory/brand/` — inside the tree HACS
copies — and fills out the set the route recognises:

- `logo.png` / `logo@2x.png`, which this PR did not produce. The wordmark is "HAventory"
  in Roboto (Bold + Light, the pairing the social preview asks a browser for), converted
  to outlines and committed so the renderer still needs no font.
- `dark_icon*` / `dark_logo*`. Without a `dark_` file Home Assistant serves the plain one
  on both themes; the two palettes are the ones in the reference artwork the owner
  supplied.
- `icon.png` moves from `#5CA1F2` to the deeper `#1F63C4`, and the pale blue — still read
  from `social-preview.html`'s `fill` — becomes `dark_icon.png`. Verified on the dev
  instance that Home Assistant asks for each on the matching theme.
- `icon.svg` is dropped; the test that compared it to the card's constant is replaced by
  re-rendering `icon.png` in the suite and comparing byte for byte.

**What this does *not* change:** the brands CDN still backs the **HACS store listing**,
which is read before anything is installed, so `ignore: brands` stays in CI and
[#196](https://github.com/chrreiter/HAventory/issues/196) still owns the decision. Filing
there is now optional rather than the only route — and it still goes to another project's
repository under the owner's name, so no session files it. The `gh` sequence in the body
above still works if you decide to; copy from `custom_components/haventory/brand/`.

The two conclusions from this PR that stand unchanged: the mark has one source (the
card's TypeScript constants) and the artwork is a rendering of it, and nothing here is
hand-drawn.

Issue #230, opened 2026-08-03
Backend subtraction: shrink the machinery to the domain


chrreiter:
### What &amp; why

The 2026-08-03 review's core backend verdict: not slop, but disciplined over-engineering — ~1,450 production LOC (~28%) removable with behavior unchanged or improved. The repo's own benchmarks set budgets at a 2,000-item typical scale that a plain scan meets, while every mutation pays the real cost elsewhere (full re-serialize per write, #200). The cut list, in priority order:

1. **Word/prefix/trigram text-index machinery** (~300 LOC: `repository.py` token cache, path-delta updates, `_search_by_text`) — removing it *fixes #220* as a side effect; `filter_items` alone meets every budget. Keep the cheap tag/category/location buckets.
2. **Index-health subsystem** (`ws.py:829-974`, ~120 LOC) — exists only to detect drift between the nine indexes; shrinks or dies with them.
3. **Debounced-persist path** (`storage.py:297-363`, ~90 LOC + 429-line test file) — called by no production mutation path. Decide together with #200: either wire it into bulk operations as #200 proposes, or delete it. Don't keep it dead.
4. **Collapse the 12 WS mutation handler tails** onto `_execute_item_op` + one serialize/broadcast/persist/reply wrapper (~250 LOC) — pairs naturally with the #221 ordering fix.
5. **One `to_dict()` per model** replacing the three hand-written item serializers (`ws.py:1779`, `repository.py:1617`, `import_export.py:90`) + the inline location dict in `ws_location_tree`.
6. **Stub-compat branches out of production** (`__init__.py:247-275` reads the offline stub's private registry; `ws.py:193-197` + ~20 `getattr` probes; `services.py:394-396`; `areas.py:25-29`) — make the stub conform to the real interface instead.
7. **Rate limiter**: shrink toward one global command bucket + minimal options (241 LOC + 9 options + 602-line test file today, off by default, LAN, authenticated). **Consolidates #199** (closed into this): the refund-on-global-reject question disappears with the second bucket; the card-side leftovers (retry-after hint in `Store.run()`, banner exclusivity, `Store.dispose()` wiring) get decided here against whatever limiter shape survives.
8. `stale_files.py` guards an empty `RETIRED_PATHS = ()` — defer the machinery until a release actually retires a file, or reduce to ~25 lines.
9. Stop persisting `_generation` (write-only in the store).

[...]

PR #618, opened 2026-08-23
test: make the offline Home Assistant stub conform, and take the stub-compat branches out of production

chrreiter:
## Summary

The offline Home Assistant stub differed from real Home Assistant in six places, and
production code carried branches for each difference. This makes the stub conform and
takes those branches out.

**Stub side** (`tests/conftest.py`, `tests/ws_helpers.py`)

- The WebSocket command registry is a dict keyed by command type, so re-registration
  replaces a handler the way HA's does instead of adding a second one.
- `area_registry.async_get` is synchronous, as HA's is.
- `RecordingConn` carries HA's own `subscriptions` registry, the disconnect that drains
  it, and core's `unsubscribe_events` pop-and-call.
- `HomeAssistant` gains a service registry that records what is registered and what is
  called, an `http` carrying aiohttp's refusal of a second route on one path, and
  `config_entries.async_forward_entry_setups` / `async_unload_platforms`.

**Production side**

- `_cleanup_ws_test_stub_registry`, `_forget_registration_flags`, the `ws_registered` /
  `services_registered` / `ws_handlers` flags: all three existed because the stub's
  registry was a list. `ws.HANDLERS` is now the command list, and the guard test reads it.
- `async_setup`, which only re-seeded a bucket `async_setup_entry` seeds itself.
- `ws.py`: the `send_message` probes, the `async_send_message` branch (reachable from
  nothing), the `on_close` / `add_close_callback` fallback, the `subscriptions` probes.
- `services.py`'s `hasattr(hass, "services")` early return; `events.py`'s bus probes;
  `todo_bridge.py`'s four entry/bus probes; `__init__.py`'s dead `getattr` / `callable`
  guards, including the two platform wrappers that existed only to hold one.
- **`areas.py` as a file**: the whole module was a wrapper whose one job was an
  `inspect.isawaitable` branch for the stub's coroutine `async_get`. `ws.py` calls
  `ar.async_get(hass)` directly and `RETIRED_PATHS` gains its first entry.

**Tests**

The stand-ins that existed for the old stub go with it: two `_ConnStub`s, `_HAConnStub`,
two `_Services`, `HassStub` / `HttpStub`. `test_integration_lifecycle_offline.py`'s
cleanup assertion becomes "a second `setup_entry` leaves one handler per command and one
service per name". The harness's own tests are retired — `ws_send` runs in 28 files and
`mountComponent` in every card spec, so a broken helper fails hundreds of tests before
its own file gets there — after moving the two cases in
`test_ws_schema_validation_offline.py` that were about shipped commands rather than
about the stub into `test_ws_items_offline.py`.

CLAUDE.md's "the offline stub has no service registry" sentence is now false and says
instead that registration is asserted offline while dispatch stays phacc-only.

Refs #230 (item 6), BE-LIFE-L2, BE-WS-C4, TESTS-T1/T7

## Counting

| | lines |
|---|---|
| production removed | 299 |
| test removed | 773 |
| added | 438 |

(`git diff --stat origin/main`: 28 files changed, 438 insertions, 1078 deletions.)

## Type of change

## How was this tested?

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` (1319 passed, 26
      skipped), `uv run ruff check .`, `uv run ruff format --check .`, `uv run mypy`
      (25 source files) — all green before every commit.
      `npx vitest run` (68 files, 1924 tests), `npm run build`.
- [x] phacc (`scripts/test_integration.sh`, card built first): **`153 passed in 22.00s`**.

The offline suite passing is not on its own evidence here — the stub is what changed —
so the phacc run is the oracle for every shape this PR now assumes: HA's keyed command
registry, `hass.services.async_register` overwriting, `entry.add_update_listener`,
`config_entries.async_forward_entry_setups` / `async_unload_platforms` /
`async_update_entry`, `ar.async_get` being synchronous, and
`ActiveConnection.subscriptions` being the only teardown path.

## Checklist

- [x] Tests added/updated — the lifecycle assertion is rewritten against what real HA
      does; nothing else was rewritten to pass.
- [x] Docs updated where behavior changed — CLAUDE.md only; no user-facing behaviour
      moved.
      `docs/data_shapes.md` in sync — no schema, error code or field moved, so neither
      document changes.

## Follow-ups

- `storage.py`'s `getattr(self._store, "key", STORAGE_KEY)` and `ws.py`'s
  `getattr(runtime.store, "schema_version", None)` are the last two probes TESTS-T1
  lists; both attributes exist on HA's `Store` and on `DomainStore`, so both are dead.
  Not in this package's list, so not touched here. Not filed — a sentence's worth.
- `rate_limit.py`'s `isinstance(getattr(conn, "subscriptions", None), dict)` is the same
  probe `ws.py` just lost. Left alone because §6.M2 deletes the file. Not filed.
- `location/update` passes `area_id` to `async_get_area` as the frame carried it — a
  `dict` or a `list` there raises `TypeError` inside the registry lookup and comes back
  as `unknown_error` instead of `validation_error`. Pre-existing, and §6.M1 PR 6
  (BE-WS-C6) is where the type answer belongs. Not filed — no real client sends it.
- `_ConnNoSend` / `_ConnRaise` in `test_ws_guard_offline.py` now exercise `_send_error`'s
  `try`/`except` rather than the probe that used to precede it. Left as they are: what
  they assert — the envelope reaches the caller even when the connection cannot take it
  — is still true and still worth pinning.

## Validate locally

1. `[dev-ha]` Deploy this branch and reload the entry twice from Settings → Devices &
   services → HAventory → Reload. Expected: both reloads report success, the sidebar
   entry stays, and the card still loads on a dashboard.
2. `[log]` `grep -E "Overwriting|Traceback" home-assistant.log` after those two reloads.
   Expected: nothing — a WebSocket command, a service and the panel each re-register
   over the top rather than raising.
3. `[browser]` With the panel open in two tabs, edit an item in one. Expected: the other
   repaints. Then close one tab and edit again in the other. Expected: no error in the
   remaining tab's console — the subscription teardown now runs through HA's own
   `subscriptions` registry on both the disconnect and the `unsubscribe` paths.
4. `[HA settings]` Developer Tools → Actions → `haventory.item_create` with
   `name: Screwdriver`. Expected: it answers, and the item appears — the service
   registration path changed shape.

## Handover

1. **Merged / left open** — this PR is for the M1 master to merge once CI (including the
   `integration` job) is green and the reload check in "Validate locally" step 1–2 has
   run on its own HA.
2. **Test this by hand** — nothing `[phone]`, `[German]` or `[owner]`-specific: no
   user-facing text, layout or stored shape moved.
3. **Validate locally** — the four steps above.
4. **Decisions taken against drifted notes** —
   - BE-WS-C4 sizes `_send_error` as fifteen lines going to one; the package names "the
     `_send_error` probe". The `getattr`/`callable` probe is gone and the `try`/`except`
     stays, because a connection that cannot be written to must not swallow the error
     envelope the caller is about to be handed. Same in `_send_event_message`: one dead
     connection must not stop a broadcast reaching the others.
   - TESTS-T1 lists `_Services`×2 for deletion, and the second one records outbound
     `todo.*` calls rather than registrations. The stub's service registry therefore
     models both directions — `async_register` and `async_call`, plus the `refuse` set a
     list that cannot delete its own lines needs.
   - TESTS-T1 also lists `storage.py:196-197` and `ws.py:867`; the package's production
     list does not, so they stay (Follow-ups).
   - `tests/integration/test_config_entry.py` lost one line asserting a flag this PR
     deletes. That file is not in the package's list, but the assertion could not
     survive the flag.
5. **Follow-ups** — the four above, none filed; none clears the real-world bar.
6. **State left behind** — branch `claude/v0-8-0-m1-stub-conformance`, five commits, no
   assets branch. Every backend PR after this one starts from the conforming stub:
   `hass.services`, `hass.http`, `hass.executor_jobs`, `config_entries.forwarded` /
   `.unloaded` and `RecordingConn.subscriptions` are all assertable now, and
   `hass.data["__ws_commands__"]` is a dict. Counting for this PR: 299 production lines
   removed, 773 test lines removed, 438 added.

---


chrreiter:
**L1 (2026-08-29) — this PR's "Validate locally", walked on the dev HA** (`main` at `94025f4` deployed over a 0.6.0 install on a wiped, re-seeded 558-item household; HA 2026.7.4; 0.7.1 control on :8124).

| step | result |
| --- | --- |
| 1 `[dev-ha]` reload the entry twice | **PASS** — both `POST …/reload` answer `{"require_restart": false}`, `haventory/version` answers after each, `get_panels` still lists `haventory`, the card loads (50 rows). The first boot of `main` over the 0.6.0 install also swept `areas.py`, `health.py`, `rate_limit.py` (`op=sweep_stale_files paths=['areas.py', 'health.py', 'rate_limit.py']`). |
| 2 `[log]` `Overwriting\|Traceback` after the reloads | **PASS** — nothing; the only ERROR in the window was the lifecycle probe's deliberate schema-108 refusal a minute earlier. |
| 3 `[browser]` two tabs, edit in one, close one, edit again | **PASS** — `haventory.item_update` with a `location_id` from a third connection repaints both tabs with the new path (`Küche` → `Garage`) and the frame each tab receives is `items/moved`; the entry reload with both tabs open shows the paused notice, then both re-subscribe and a fresh create repaints both; no console error. |
| 4 `[HA settings]` `haventory.item_create` | **PASS** — `call_service` with `return_response` answers `{"item": {…}}`; the item appears. |

Evidence: [`m1/01-card-after-reloads.png`](https://raw.githubusercontent.com/chrreiter/HAventory/claude-assets-v0-8-0-l1/l1/m1/01-card-after-reloads.png), [`m1/06-two-tabs-moved.png`](https://raw.githubusercontent.com/chrreiter/HAventory/claude-assets-v0-8-0-l1/l1/m1/06-two-tabs-moved.png), [`m1x/619-4-after-reload.png`](https://raw.githubusercontent.com/chrreiter/HAventory/claude-assets-v0-8-0-l1/l1/m1x/619-4-after-reload.png). The full L1 record is on #236.

PR #620, opened 2026-08-23
refactor(ws): one tail for every single-item mutation

chrreiter:
## Summary

Eleven single-item WebSocket commands each carried their own copy of the same
tail — run the operation, persist, announce, reply. Six of those copies were
inline twins of the `_op_item_*` entry `haventory/items/bulk` dispatches to, and
five more rebuilt a payload dict out of the frame they had just been handed.
`_mutate_item` is that tail once; each handler is now its schema, its guard and
one line, and a command and the bulk row naming the same `kind` reach the
repository through one function.

`media.async_delete_item_files` moves into that tail, so the rule #592 put in
one place is now reached from one place on the WebSocket side, and the test its
follow-up asked for lands with it: all three delete surfaces asserted against
that one call, once, instead of a test per caller.

`ws_items_bulk`'s per-row error mapping was a hand-built version of what
`ws_guard` already does for a whole command; it now uses the same two pieces.

Refs #230 (items 4 and 6 of its 2026-08-22 comment: BE-WS-C1, BE-WS-C7), refs #565.

## Two answers move, and what they are

Both are the collapse making a command answer the way its bulk row already
answers — or the other way round where the model's answer is the contract.

| Payload | Before, `item/*` | Before, `items/bulk` | After, both |
|---|---|---|---|
| `{item_id: <unknown id>, quantity: -1}` | `validation_error` | `not_found` | `validation_error` |
| `{item_id: 7, name: "X"}` (id not a string) | `not_found` | `validation_error` | `validation_error` |

The first is the asymmetry the audit recorded: `ws_item_set_quantity` pre-checked
`quantity < 0` with a message copied from `models.py`, and `_op_item_set_quantity`
did not check at all, so the batch looked the item up first. The rule now lives in
`models.validate_quantity` — one function replacing the three copies create, update
and the core-field check each carried — and the op runs it before the lookup.

The second follows from the same ordering: `_payload_item_id` refuses a
non-string id, which the eleven commands did not do for themselves. Both are
what `docs/backend_api_contract.md` already says about an `object`-typed field:
the handler names the value rather than letting the id decide the code. The
contract gains one sentence saying it for the pair of surfaces at once;
`docs/data_shapes.md` needed nothing — no shape moved, only which code answers a
value the shape already forbade (`quantity?: number>=0` was and stays the
documented type).

## Counting

| | Lines |
|---|---|
| Production removed (`custom_components/`) | 198 |
| Test lines removed | 1 |
| Lines added | 278 |

`git diff --stat` against `main`: `models.py` +23 −14, `ws.py` +103 −184,
`backend_api_contract.md` +1, `scripts/stress_test.py` +3 −1,
`test_delete_surfaces_offline.py` +84, `test_item_input_validation_offline.py`
+64 −1. Production net −72; the added test lines are the delete-surface
assertion and the two answers-alike cases.

## Type of change

## How was this tested?

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` → **1325
  passed, 26 skipped** (5 new; `main` collects 1320). `uv run ruff check .` → All
  checks passed. `uv run ruff format --check .` → 187 files already formatted.
  `uv run mypy` → Success: no issues found in 26 source files.
- [x] In-process (phacc), card built first: `bash scripts/test_integration.sh` →
  **153 passed in 20.23s** — `[ OK ] Integration tests OK`.
- [x] Frontend (`cards/haventory-card`): `npx eslint .` clean, `npm run typecheck`
  clean, `npx vitest run` → 68 files, 1924 passed, `npm run build` → 745.77 kB.
  No card file is touched; run because the card is built for phacc anyway.

Each of the four commits was gate-green on its own. The two behaviour commits
were seen failing first: `test_a_refused_quantity_answers_alike_from_the_command_and_from_bulk`
fails with `assert 'not_found' == 'validation_error'` before the quantity fix,
and `test_an_item_id_of_the_wrong_type_names_the_field_on_both_surfaces` fails
the same way with `ws.py` alone stashed. The delete-surface test was seen failing
by dropping the `async_delete_item_files` call out of the tail, which reddens the
`item/delete` case and leaves the other two green — the shape of the bug #592
fixed.

The eleven collapsed handlers needed no test change at all, which is what says
the cut is a cut: `test_ws_items_offline.py`, `test_ws_bulk_offline.py`,
`test_ws_error_mapping_offline.py` (which monkeypatches `_op_item_update`, so the
dispatch table stays resolved per call), `test_ws_guard_offline.py` and
`test_ws_logging_offline.py` all pass untouched.

New tests:

- `tests/test_delete_surfaces_offline.py` — one parametrized case over
  `haventory/item/delete`, an `item_delete` row of `haventory/items/bulk` and
  `haventory.item_delete`, each asserted against `media.async_delete_item_files`
  with the deleted body. Ordering and what a failed write frees stay with each
  caller's own tests; this one holds the rule a fourth surface would have to meet.
- `tests/test_item_input_validation_offline.py` — the two answers-alike cases in
  the table above.

## Checklist

- [x] Docs updated where behavior changed.
  `docs/data_shapes.md` in sync.

- **`_mutate_item` takes no `counts` parameter.** The plan's signature has
  `counts=True`, but none of the eleven table-shaped commands suppresses the
  counts event — the two that do (`attachment/update`, `attachment/reorder`) are
  not table-shaped and keep their own bodies. A parameter with one possible value
  is what this milestone removes, so it is not written; a later caller that needs
  it adds it with its reason.
- **The bulk row reuses `_context_from_msg` and the logging half of
  `_error_message`.** That half is now `_log_rejection`, which returns the
  `(code, message)` pair the wire carries. Calling `_error_message` itself would
  have meant building a whole error envelope under a fake frame id in order to
  read two fields back out of it.
- **The `set_quantity` rule went into `models.validate_quantity` rather than
  staying restated in `ws.py`.** The plan says the pre-check goes because it
  restates the model; moving it into the op would only have moved the
  restatement. `models.py` is edited for that one function, which replaces three
  copies of the same rule and its message.
- **`scripts/stress_test.py` is edited** (one line): it counted refused bulk rows
  by matching the fixed sentence that log line used to carry. The line now
  carries the reason as its message, so the script matches the structured
  `items_bulk_op_failed` instead.

## Follow-ups

- `scripts/stress_test.py` still counts `Persist requested, debouncing` and
  `Cancelled pending persist task`, log lines PR #617 deleted, so its debounce
  metrics are permanently zero. Not filed: the maintained stress driver is
  `.claude/skills/test-haventory/stress.py`, and this one is a dev script whose
  sweep belongs with whatever revisits the skills.
- `_payload_int` now has one caller left (`_op_item_adjust_quantity`); BE-WS-C6
  retires it along with the `object`-typing pass.
- The row error envelope calls its context `context` while every command-level
  envelope calls it `data`. One name would be better, but the card reads neither
  (it reads `code` and `message` off a row), so the change is a wire rename with
  no reader — worth doing only when something else moves that envelope.

## Handover

1. **Merged / left open** — this PR is left open for the master to squash-merge
   once CI is green; nothing here needs the owner. CI on `ada45ea`: all 12 checks
   green, `integration` and `backend (3.14)` included.
2. **Test this by hand** — nothing `[phone]`, `[German]` or `[owner]`-only: no
   card file, no string, no dictionary is touched.
3. **Validate locally**
   1. `[dev-ha]` Deploy the branch and reload the entry. `[HA settings]`
      Developer Tools → Actions → `haventory.item_delete` on an item carrying a
      picture; the file under `<config>/haventory/attachments/<item_id>/` is gone
      and the response envelope is unchanged.
   2. `[browser]` In the card, edit an item, move it, adjust and set its
      quantity, check it out and back in, add and remove a tag, and change its
      low-stock threshold; each still answers and each still repaints a second
      open tab. These are the eleven collapsed commands.
   3. `[browser]` Select several items in the bulk bar and delete them; their
      pictures are gone from disk, and a row refused on a stale version keeps
      its own.
   4. `[log]` Nothing at ERROR carrying `haventory` through any of the above; a
      refused bulk row logs at WARNING with `op=items_bulk_op_failed` and the
      reason as its message.
4. **Decisions taken against drifted notes** — the four above.
6. **State left behind** — branch `claude/v0-8-0-m1-mutation-tail`, no assets
   branch, no `.env` written, no Home Assistant instance touched. Counting for
   this PR: 198 production lines removed, 1 test line removed, 278 added.

### DIFF
--- a/custom_components/haventory/models.py
+++ b/custom_components/haventory/models.py
@@ -1149,11 +1149,27 @@ def _validate_optional_text(
         raise ValidationError(f"{field_name} must be at most {max_length} characters")
 
 
+def validate_quantity(value: object) -> int:
+    """Return the quantity, or refuse the value with the one message there is.
+
+    Public because a caller that holds an unvalidated quantity before it holds
+    an item — `haventory/item/set_quantity` and the `items/bulk` row of the same
+    kind — checks the value first, so a payload that is wrong about both is
+    answered on the value rather than on the id.
+
+    Spelled out rather than through ``_is_int_not_bool`` so the comparison and
+    the return narrow to ``int``.
+    """
+
+    if not isinstance(value, int) or isinstance(value, bool) or value < 0:
+        raise ValidationError("quantity must be an integer >= 0")
+    return value
+
+
 def _validate_item_core_fields(name: str, quantity: int, low_stock_threshold: int | None) -> None:
     if len(validate_required_name(name)) > NAME_MAX_LENGTH:
         raise ValidationError(f"name must be at most {NAME_MAX_LENGTH} characters")
-    if not _is_int_not_bool(quantity) or quantity < 0:
-        raise ValidationError("quantity must be an integer >= 0")
+    validate_quantity(quantity)
     if low_stock_threshold is not None and (
         not _is_int_not_bool(low_stock_threshold) or low_stock_threshold < 0
     ):
@@ -1185,14 +1201,10 @@ def create_item_from_create(
     # returns the trimmed value this stores.
     name = validate_required_name(payload.get("name"))
     description = payload.get("description")
-    # Type before conversion, for the same reason `name` is checked before
-    # `.strip()`: the command schema types `quantity` as `object`, and `int()`
-    # over a non-numeric string raises `ValueError`, which routes to
-    # `unknown_error` instead of naming the field.
-    raw_quantity = payload.get("quantity", 1)
-    if not _is_int_not_bool(raw_quantity):
-        raise ValidationError("quantity must be an integer >= 0")
-    quantity = int(raw_quantity)
+    # Type before use, for the same reason `name` is checked before `.strip()`:
+    # the command schema types `quantity` as `object`, and arithmetic over a
+    # non-integer routes to `unknown_error` instead of naming the field.
+    quantity = validate_quantity(payload.get("quantity", 1))
     status = validate_item_status(
         payload.get("status", DEFAULT_ITEM_STATUS), known_statuses=known_statuses
     )
@@ -1277,10 +1289,7 @@ def _update_name_and_description(new_item: Item, update: ItemUpdate) -> None:
 
 def _update_quantity(new_item: Item, update: ItemUpdate) -> None:
     if "quantity" in update:
-        q = update["quantity"]
-        if not _is_int_not_bool(q) or q < 0:
-            raise ValidationError("quantity must be an integer >= 0")
-        new_item.quantity = q
+        new_item.quantity = validate_quantity(update["quantity"])
 
 
 def _update_status(new_item: Item, update: ItemUpdate, known_statuses: Collection[str]) -> None:
--- a/custom_components/haventory/ws.py
+++ b/custom_components/haventory/ws.py
@@ -505,6 +505,33 @@ async def _persist_repo(hass: HomeAssistant) -> None:
     await storage_mod.async_persist_repo(hass)
 
 
+async def _mutate_item(
+    hass: HomeAssistant, conn: websocket_api.ActiveConnection, msg: dict[str, Any], kind: str
+) -> None:
+    """Run one item operation and answer for it: write, free, announce, reply.
+
+    The command's fields are the op's payload — the envelope's `id` and `type`
+    are all that is stripped — so a command and the `items/bulk` row naming the
+    same `kind` reach the repository through one function and answer alike.
+    """
+
+    payload = {k: v for k, v in msg.items() if k not in {"id", "type"}}
+    serialized, action = _execute_item_op(hass, kind, payload)
+    await _persist_repo(hass)
+
+    # `deleted` is the action `_op_item_delete` alone returns, so it names
+    # exactly the body whose files nothing references any more — and exactly the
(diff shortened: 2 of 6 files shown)


chrreiter:

| step | result |
| --- | --- |
| 1 `haventory.item_delete` on an item carrying a picture | **PASS** — the response envelope carries the removed body; `attachments/<item_id>/` (original + tile) is gone; a bulk `item_delete` frees the files the same way; `attachments/` itself stays. |
| 2 the eleven collapsed commands each answer and repaint a second tab | **PASS** — update, move, adjust, set quantity, check out, check in, add tags, remove tags, threshold, custom fields (`set`), reminder: each `success`, each delivered to both open tabs, and the item carried every write afterwards (v12: description, q7, Garage, threshold 1, `Farbe: rot`, reminder 2026-09-30). A tag added from the card's editor lands and its chip repaints. |
| 3 a bulk with one row refused on a stale version | **PASS** — the fresh row applied, the stale row answered `conflict` ("version conflict: expected 1, actual 13") and kept its own, and the log carries `WARNING … op=items_bulk_op_failed item_id=… expected_version=1`. |
| 4 `[log]` | **PASS** — nothing at ERROR; the refused row at WARNING with the reason as its message. |

PR #621, opened 2026-08-23
refactor(services): services and WebSocket share one op table

chrreiter:
## Summary

Every `haventory.*` service was a second implementation of a write the WebSocket surface
already had. `ops.py` now holds one function per write, and both surfaces run it.

- **`ops.py`** — the eleven `_op_item_*` entries move out of `ws.py` as they were, and are
  joined by `item_create`, `reminder_bump` and the three location verbs. Each op takes the
  payload as a plain dict, makes the repository call and returns a `Written` (the entity, the
  event it earns, whether the tree needs a repaint). `announce` is the step after the persist:
  the files a delete orphaned, then the door in `events.py`.
- **`services.py`** — twelve handlers of eight identical lines and one identical `except`
  become `_run_service(name, data)`: schema → op → persist → announce → envelope. The twelve
  `service_*` names stay as one-line entries, because they are what the `SERVICES` table binds
  and what the tests call.
- **`ws.py`** — `ws_item_create`, `ws_reminder_bump` and `ws_location_update` lose the bodies
  that duplicated a service, and `ws_location_create` / `ws_location_delete` go the same way
  since the ops existed by then. `_mutate_item` becomes `_mutate` and covers both nouns.
- The location-update "what changed decides which event" step, written twice (`ws.py` and
  `services.py`, with the comment reworded), is now written once in `_op_location_update`.
- The area-registry check the two location commands made twice is one `_require_known_area`.
  It stays on the WebSocket side: the services never made it, and making them refuse an
  unknown `area_id` would be a behaviour change this PR is not carrying (see Follow-ups).
- **`repository.py`** gains `iter_locations()` / `children_of(parent_id)`, so `location/list`
  and `location/tree` stop reaching through `_debug_get_internal_indexes()` and the private
  child map. The debug accessor keeps its other callers (`health.py`, tests) until §6.M1's
  health PR.

What survives, and what pins it: the typed service schemas (`test_services_schema_offline.py`),
`new_location_id` and the required `due_date` on check-out (same file, plus the catalog test),
the `{"item": …}` / `{"location": …}` envelope (`test_every_service_answers_with_the_canonical_envelope`),
the per-op log context (a new test over the table, plus `test_ws_logging_offline.py`),
`_bind`'s coroutine-function constraint (`test_service_registration_and_schema_errors`), and
the catalog test holding the schemas, `services.yaml` and `strings.json` to one field set.

Two commits carry a change of behaviour, each with its own test:

- **`haventory.item_update` carrying a `location_id` now announces `items/moved`**, as
  `haventory/item/update` always did. The contract already says both write paths emit the same
  events; they did not.
- **`services.yaml`: one selector per field, and an example on every one** (#593). `tags` was
  `text:` on `item_create` and `object:` on `item_update` for the value both schemas take as a
  list, so the create form in Developer Tools → Actions offered a single-line box that cannot
  produce one. Thirty-seven fields carried no `example`.

Refs #230
Refs #593

## Counting

| | lines |
|---|---:|
| production removed | 538 |
| tests removed | 8 |
| added | 729 |

`git diff origin/main --numstat`. The added figure carries `ops.py` whole (367), of which ~180
are the op layer moved out of `ws.py` unchanged and the rest is the new shared tail plus the
five ops that had no table entry; `services.yaml` accounts for 40 more (the examples). Between
them `ws.py` and `services.py` lose 380 lines net.

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue) — the two commits named above
- [x] Refactor (no functional change) — everything else

## How was this tested?

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` → **1331 passed, 26 skipped**;
  `uv run ruff check .` → all checks passed; `uv run ruff format --check .` → 188 files already
  formatted; `uv run mypy` → success, 27 source files (`haventory.ops` joins the strict list).
  → **68 files, 1924 tests passed**, `npm run build` → built in 192ms. Untouched by this PR;
  run because the gate says so and because phacc needs the bundle.
- [x] phacc (`scripts/test_integration.sh`, card built first): **153 passed in 20.37s** —
  `[ OK ] Integration tests OK`.

New or changed tests:

- `test_service_broadcast_offline.py` — an update carrying a location announces the move on
  both surfaces (red before the fix).
- `test_services_offline.py` — every service names the fields its refusal logs; a refused
  `item_move` carries `new_location_id` and a refused `item_create` carries `item_name`; every
  yaml field shows an example and a field name declares one selector across services.
- `test_repository_locations_offline.py` — `iter_locations()` / `children_of()`, including that
  the answer is not the live index.
- `test_docs_contract_offline.py` — the contract's `items/bulk` kinds are exactly `ops.BULK_KINDS`,
  and that set is a strict subset of the op table. Verified red by removing one name from the
  document.
- `test_ws_error_mapping_offline.py` — two monkeypatch targets follow the symbols that moved
  (`ws.serialize_item` → `ops.serialize_item`, `ws._op_item_update` → `ops.OPS["item_update"]`).
  The assertions are unchanged.

## Checklist

- [x] Docs updated where behavior changed — the contract's "one payload, one answer" guarantee
  now names the third surface; `CLAUDE.md`'s module map gains `ops.py`.
- [x] WebSocket API changes keep `ws.py`, `docs/backend_api_contract.md` and `docs/data_shapes.md`
  in sync — no schema, code or field moved, so `data_shapes.md` is unchanged.
  optimistic `version`).

## Follow-ups

- **`haventory.location_create` / `location_update` accept an `area_id` Home Assistant has no
  area for**, where the commands refuse it: a location anchored to a typo reports an
  `effective_area_id` that filters to nothing, and nothing says so. The op is shared now, so
  the fix is one call moved into it. Not filed: it needs a decision on whether an automation
  that has been storing such an id should start failing, which belongs with the owner.
- **`ws_location_move_subtree` still announces `moved` for a move that changed no parent**,
  where `location/update` announces nothing. Left alone: the contract documents the current
  answer, so changing it is a contract change rather than a cut. Not filed.
- A service call with an empty-string `item_id` now answers `validation_error` ("item_id must
  be a non-empty string") rather than whatever the repository made of `""`, and
  `item_set_quantity` answers on the quantity before it looks the item up — both are the
  answers the command and the bulk row already gave. Named here rather than filed.

## Handover

1. **Merged / left open** — this PR is left open for the master to review and squash-merge; it
   is not on the owner-merges list.
2. **Test this by hand** — `[owner]` Developer Tools → Actions renders `haventory.item_create`'s
   `tags` as an object field with an example, and every other field shows one. No `[phone]` or
   `[German]` half: no dictionary key moved.
3. **Validate locally** — see the numbered block in the report; nothing user-visible beyond the
   two named behaviour changes.
4. **Decisions taken against drifted notes** — the op table lives in a new `ops.py` rather than
   in `ws.py` with `services.py` importing it, because `serialization.py` exists precisely so
   the two surfaces need not import each other. BE-WS-C8's "put the decision beside
   `Repository.update_location` or in `events.py`" is answered by the op instead: it is a write
   plus its consequence, and `events.py` is doors only.
6. **State left behind** — branch `claude/v0-8-0-m1-services-op-table`, no assets branch, no HA
   instance touched. `_debug_get_internal_indexes` still has `health.py` and the tests as
   callers, as §6.M1's PR 8 expects.

---


chrreiter:
**L1 (2026-08-29) — the service half of M1's "Validate locally", walked on the dev HA** (`main` at `94025f4`, 558-item seeded household, HA 2026.7.4).

| step | result |
| --- | --- |
| `haventory.item_create` from Developer Tools → Actions | **PASS** — `get_services` shows all 14 fields with an `example`, `tags` as an `object` selector (the same selector on `item_update`), `response: {optional: true}`; a `call_service` with `return_response` answers `{"item": {…}}`. |
| `haventory.item_update` with `tags: kitchen` (a string) | **PASS, with a note** — refused and the item unchanged (version and tags identical afterwards). The refusal is core's `invalid_format: expected a list for dictionary value @ data['tags']` from the typed service schema, not the model's `validation_error` — the service schemas kept their types (§6.M1 PR 5), so a wrong type never reaches the model on the service path. |
| `haventory.item_delete` on an item with a picture | **PASS** — the response carries the removed body; `attachments/<item_id>/` (original + tile) is gone; a bulk `item_delete` frees the files the same way; `attachments/` itself stays. |
| `haventory.item_update` with a `location_id`, two tabs open | **PASS** — both tabs repaint with the new path and receive `items/moved`. |
| The services' German names and descriptions | rows in L1's #540 table, for the owner's read. |

Evidence: [`m1/04-developer-tools-item-create.png`](https://raw.githubusercontent.com/chrreiter/HAventory/claude-assets-v0-8-0-l1/l1/m1/04-developer-tools-item-create.png). The full L1 record is on #236.

PR #622, opened 2026-08-23
refactor(models): type every payload field the same way

chrreiter:
## Summary

The sixth PR of §6.M1, carrying `BE-WS-C6`. Every collection a caller writes whole is typed
`object` in its command schema, so the model answers a wrong type with `validation_error`
naming the field, instead of Home Assistant refusing the frame as `invalid_format` before the
guard runs — which logs the client's payload at ERROR and names nothing.

`tags` was the field on both sides of that split: concrete on `item/create`, `item/add_tags`
and `item/remove_tags`, `object` on `item/update`, so the same wrong value answered two
different codes depending on which command carried it. #591's guard under `validate_tags` is
what makes closing that safe — a non-list and a non-string entry are both refused rather than
iterated as characters.

Refs #230

### What changed on the wire

| Command | Field | Was | Is |
|---|---|---|---|
| `item/create` | `tags`, `custom_fields` | `invalid_format` | `validation_error` |
| `item/update` | `custom_fields_set`, `custom_fields_unset` | `invalid_format` | `validation_error` |
| `item/add_tags`, `item/remove_tags` | `tags` | `invalid_format` | `validation_error` |
| `item/update_custom_fields` | `set`, `unset` | `invalid_format` | `validation_error` |
| `item/attachment/reorder` | `attachment_ids` | `invalid_format` | `validation_error` |
| `status/reorder` | `slugs` | `invalid_format` | `validation_error` |
| `item/list`, `export`, `distinct_values`, `location/tree` | `filter.tags_any`, `filter.tags_all` | an empty page | `validation_error` |
| `subscribe` | `location_ids` | an entry coerced with `str()` | `validation_error` |

All four filter surfaces were checked by hand against a seeded repository, not assumed.
Nothing is written under either code, before or after. A client that already handled both —
the card does: `host-surfaces.ts` reads `validation_error`'s field errors and everything else
is the generic banner — needs no change, which is what keeps the one restart of an in-place
upgrade safe in both directions (a new bundle against the old backend still sees
`invalid_format`, and handles it).

### The rest

- `models.require_string_list` is the type check `normalize_string_list` already made, without
  the trimming and de-duplication: `status/reorder` and `item/attachment/reorder` name a whole
  set as a permutation, and normalizing first would accept a list naming one member twice.
  `normalize_string_list` is now written in terms of it.
- `custom_fields_set` / `custom_fields_unset` are read by presence rather than by truthiness,
  so a non-mapping that happens to be falsy is refused rather than skipped.
- `selected_tags` puts `tags_any` / `tags_all` on the rule `categories` and `location_ids`
  already follow. Both the scan and the index pre-filter read through it — the pre-filter runs
  first, and would otherwise return the empty page before the scan could refuse anything.
  `normalize_tags` keeps its tolerant coercion for its one remaining caller, an import
  document, where a restore must not fail on a row an earlier release wrote.
- `_payload_int` restated `Repository.adjust_quantity`'s own check, in the same words and with
  the same message, for its one caller. The repository is where it has to be — `delta` reaches
  the arithmetic through that method from every surface — so the parameter is typed `object`
  and narrowed by the guard that was already there.
- The contract's paragraph enumerating the fields that keep a concrete schema type is gone, and
  with it the half of `tests/test_docs_contract_offline.py` that held that list to the code. A
  swept behaviour test replaces it: every collection field, given a bare string, answers
  `validation_error` and writes nothing.

### Counting

| | Lines |
|---|---|
| Production removed (`custom_components/`) | 57 |
| Tests removed | 163 |
| Docs removed | 43 |
| Added | 291 |

`git diff --stat origin/main`: 14 files changed, 291 insertions(+), 263 deletions(-). This PR
is not a subtraction on balance — the list-maintenance test it deletes is replaced by a swept
behaviour test twelve cases wide, and four of the widened fields had no model-side type check
at all.

## Type of change

## How was this tested?

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` → **1356 passed, 26 skipped**;
      `uv run ruff check .`, `uv run ruff format --check .` (188 files), `uv run mypy`
      (27 source files) → clean
      `npx vitest run` → **1924 passed / 68 files**, `npm run build` → 745.77 kB
- [x] phacc (`scripts/test_integration.sh`, card built first) → **153 passed in 19.26s**

New and changed tests:

- `tests/test_item_input_validation_offline.py` — eleven parametrized cases, one per
  collection field under a command that takes it: a bare string answers `validation_error`,
  the message names the caller's own key, and neither the item nor the count moves. Plus the
  edge case the type check must not swallow: `status/reorder` still refuses
  `["ok", "ok", "missing"]`.
- `tests/test_ws_items_offline.py` — `tags` on all three write commands answers one code with
  one message; `item/list` refuses a bare-string `tags_any` / `tags_all` rather than answering
  an empty page.
- `tests/test_models_filter_sort_offline.py` — the existing multi-select parametrization gains
  `tags_any` and `tags_all` (five bad values each).
- `tests/test_ws_subscriptions_offline.py` — five bad `location_ids` values, each refused with
  the field named and no subscription registered.
- `tests/integration/test_ws_items.py` — three frames over the real voluptuous schemas, which
  is the only mode that can tell "the schema let it through and the model answered" apart from
  "the schema never typed it".

## Checklist

- [x] Docs updated where behavior changed (`docs/backend_api_contract.md`, `docs/data_shapes.md`).
- [x] WebSocket API changes keep `ws.py`, `docs/backend_api_contract.md`, and
      `docs/data_shapes.md` in sync.
      optimistic `version`).
- [x] No file removed from `custom_components/haventory/`, so `RETIRED_PATHS` is untouched.
- [x] No dictionary edited (`strings.json`, `translations/*.json`, `src/i18n/*.ts`).

None — no card file changed.

## Follow-ups

- **The simplified rule is not yet true for the scalars.** §6.M1 PR 6 words the replacement as
  "`validation_error` is the model's answer to every *value* — a wrong type included". The
  package's own scope is "the remaining `[str]` / `{str: object}` fields", which is what this
  PR did, so about twenty scalars still answer `invalid_format`: `expected_version`, the dates
  and `reminder_interval`, `checked_out`, the attachment handles, `haventory/subscribe`'s
  `topic` and two flags, the `haventory/status/*` scalars, and import's `document` / `policy`.
  The contract now names them and says the list is not itself the contract, rather than
  claiming the stronger rule. Finishing it is not only a widening — four of those have no
  model-side check to fall back on, and three more (`checked_out`, `include_subtree`,
  `inspection_overdue_only`) are coerced with `bool()` today, so each would be its own
  behaviour change with its own commit. **Not filed**: it is a package decision for the master,
  not a defect a household can hit.
- **`validate_custom_fields`' cap messages still say `custom_fields`** when a patch was sent
  under `custom_fields_set`. Deliberate — the caps bound the item's map, not the patch — and
  only the shape refusal takes the caller's key. **Not filed.**
- **The service schemas keep their concrete types**, which §6.M1 PR 5 lists as what must
  survive. A `haventory.item_update` carrying `tags: kitchen` answers `validation_error`
  already (`services.py` grades a `vol.Invalid` as one), so the two surfaces agree on the code
  and differ only in the message text. **Not filed.**

## Handover

**Merged / left open** — this PR is open; the M1 master merges it after review. Nothing stacks
on it inside this package: PR 7 touches `repository.py`'s search path, which this PR only edits
at the `tags_any` pre-filter line PR 7 deletes outright.

**Test this by hand** — nothing `[phone]` or `[German]`. `[owner]` only if a household drives
the API from a script: a script that sent `tags` as a bare string used to get `invalid_format`
and now gets `validation_error`; both are refusals and neither writes, so nothing to repair.

**Validate locally**

1. `[dev-ha]` Deploy the branch and reload the entry from Settings → Devices & services.
2. `[HA settings]` Developer Tools → Actions → `haventory.item_update`, `tags: kitchen` (a bare
   string, not a list). Expected: `validation_error`, the item unchanged. This is the §6.M1
   step; it answered the same code before this PR and must keep doing so.
3. `[browser]` With the card open on an item, edit its tags and save. Expected: the edit lands
   and the chips repaint — the card sends a list from every path, so nothing here should have
   moved.
4. `[log]` Nothing at ERROR carrying `haventory` through steps 2 and 3. A wrong-typed
   collection now logs at WARNING with the field named, where core used to log the client's
   whole payload at ERROR.

**Decisions taken against drifted notes**

- C6's blocker ("land the `validate_tags` list guard first") was already resolved on `main` by
  #591; nothing was added for it.
- C6 names `tests/test_ws_schema_validation_offline.py:229-238` as pinning `tags="chisel"` to
  the HA-core rejection. PR 2 moved that case out; it lives in `test_ws_items_offline.py` and
  is what this PR rewrites.
- The `qty < 0` pre-check C6 lists beside `_payload_int` was already gone — PR 4 took it.
- The plan's wording for the replacement rule was written for a full widening. Scoped to what
  the package names, and the contract says what is true; see Follow-ups.

**State left behind** — branch `claude/v0-8-0-m1-uniform-typing`, no assets branch, no Home
Assistant instance touched. Counting for this PR: 57 production lines removed, 163 test lines,
43 doc lines, 291 added.


chrreiter:

| step | result |
| --- | --- |
| 1 deploy, reload the entry | **PASS** — reloaded twice from the REST endpoint Settings uses. |
| 2 `haventory.item_update` with `tags: kitchen` (a bare string) | **PASS, with a note** — refused, item unchanged (version and tags identical). The answer is core's `invalid_format: expected a list for dictionary value @ data['tags']` from the typed service schema (kept on purpose by #621), not the model's `validation_error` — the wrong type never reaches the model on the service path, so the log carries no WARNING naming the field for it either. Over WebSocket the same value reaches the model and answers `validation_error` (the fuzz layer's `qty bool`/`cf nonscalar` cases pass the same way). |
| 3 edit an item's tags in the card and save | **PASS** — the tag lands and the editor's chips repaint. |
| 4 `[log]` | **PASS** — nothing at ERROR through it. |

PR #624, opened 2026-08-23
refactor(repository): the scan answers q; delete the text index

chrreiter:
## Summary

`q` was answered twice. `_get_filtered_candidates` consulted a word / name-prefix / trigram
index first, and then `filter_items` applied `_item_matches_q` to every item that survived — so
the buckets could only ever narrow the answer the scan would have given, never widen it. This
deletes them: the buckets, the per-item token cache, the path-token delta that kept them in step
through a subtree move, and the `q` step of `_get_filtered_candidates`. The cheap tag / category /
location / status / area buckets stay exactly as they are.

One shape of that narrowing was wrong, and is fixed here — see "The behaviour half" below.

Closes #623
Refs #230 (item 1, BE-REPO-1)

## Type of change

## The measurement, before and after

The 0.5 s budget the issue names came from a benchmark test that is no longer in the tree, so the
first commit puts one back: `tests/test_repository_search_benchmark_offline.py` seeds 10,000 items
across 20 locations — name, description, category, tags and a denormalized display path each
populated — and times `list_items(flt={"q": "Widget"})`, which returns a fifth of them. It is
opt-in with `ASSERT_BUDGETS=1`, because a wall-clock assertion on a shared runner fails for
reasons that have nothing to do with this code.

```
ASSERT_BUDGETS=1 PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q -s \
    tests/test_repository_search_benchmark_offline.py
```

| | best of 5 passes | budget |
|---|---|---|
| before the cut (index answers `q`) | **0.010 s** | 0.5 s |
| after the cut (scan answers `q`) | **0.020 s** | 0.5 s |

Twice the cost of the indexed path, 25x under the ceiling, at five times the largest inventory
the release testing plan drives. The write side moved the other way: the same test file's total
wall clock, which is dominated by seeding those 10,000 items, went from 1.86 s to 0.81 s, because
a create or an update no longer tokenizes five fields and touches three bucket maps. The offline
suite went from 6.6 s to 4.6 s for the same reason.

## The behaviour half

The cut is not behaviour-neutral in one direction, and the direction is a fix.

`_get_candidates_for_word` reached a mid-word match only through its trigram fallback, and ran
that fallback **only when the word bucket and the name-prefix bucket both missed**. So as soon as
any one item had the query as a whole word or as the start of a name word, every *other* item
carrying the query inside a word was dropped — and `_text_index_covers_query` called a
three-character query covered, so the short answer was treated as authoritative and the scan never
saw the missing items.

An inventory holding **Light** and **Flashlight**: `q="light"` returned Light alone. Delete the
item called Light and the same query returns both. Which item is in the store decided whether
another item was findable. Same for Screw/Unscrew, Case/Suitcase, Box/Toolbox. Filed as #623 and
closed by this PR; `test_mid_word_match_survives_a_word_start_hit_elsewhere` is the pin, and it is
the one new test that fails on `main`.

This is the second half of what #230 item 1 predicted — #220 fixed the sub-trigram fragments in
0.3.1 by letting them fall through to the scan; this removes the short-circuit for the rest.

## Tests

The five fall-through tests from #220's fix are the oracle, because they ask `list_items` what it
returns rather than asking the index what it holds. Four of them keep every assertion they had and
pass unchanged over both trees; two of those four had a docstring naming a bucket, rewritten to
state the contract instead, and were renamed for the same reason. What went:

- the six `_get_filtered_candidates({"q": …})` assertions — they name a pre-filter that no longer
  exists. `test_short_fragment_falls_through_while_long_one_stays_indexed` held four of them and
  nothing else, so the fifth fall-through test goes with them; the other two keep their
  `list_items` assertions and lose only the private-accessor half.
- `test_text_indexes_match_fresh_rebuild_after_moves` — the three-dict comparison against a
  rebuilt repository, with no dicts left to compare. What it was really guarding, that search
  agrees with the display path after a move and a rename, is still pinned by the two tests above
  it in that module.

What is new, all of it asserting results:

- `test_mid_word_match_survives_a_word_start_hit_elsewhere` — Light / Flashlight, above.
- `test_accented_query_matches_unaccented_content` — `q="café"` finds an item called `Cafe`. The
  existing test covered the other direction only.
- `test_multi_word_query_ands_across_fields_and_the_path` — one query word each in name,
  description, tag, category and the location display path; and one word no field carries drops
  the item.
- `test_a_query_that_normalizes_away_narrows_nothing` — a Japanese item name (`電池ケース`) with
  `電池` as the query. `normalize_search_text` keeps only what NFKD renders as ASCII, so the query
  reduces to no words, and a filter with no words to test excludes nobody: every item comes back.
  Identical before and after the cut — the scan always answered this one — and written down
  because it is the contract for a script without word boundaries. See Follow-ups.

Docstrings that named the buckets were rewritten to state the contract instead, in
`repository.py`, `models.normalize_search_text`, and the two test modules.

## Counting

| | lines |
|---|---|
| production removed (`custom_components/`) | 260 |
| test lines removed | 79 |
| lines added | 203 (95 the opt-in benchmark, 92 in the two rewritten test modules, 16 production and docs) |

`repository.py`: 2,700 → 2,452 lines.

## How was this tested?

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` → **1358 passed, 27 skipped**;
  formatted; `uv run mypy` → Success: no issues found in 27 source files.
  (**68 files, 1924 passed**), `npm run build` — untouched by this PR, run because the card is
  built before phacc.
- [x] phacc (`scripts/test_integration.sh`, card built first): **153 passed in 20.35s**.

## Checklist

- [x] Tests added/updated (happy path + edge cases).
- [x] Docs updated where behavior changed — `docs/data_shapes.md`'s `q` line now states the
  contract the scan implements (substring, mid-word included, and `category` among the fields it
  reads, which it always did).
- [x] No WebSocket schema, error code or field moves here.
- [x] Essential invariants preserved: search stays case- and accent-insensitive (and gains the
  mid-word matches it promised), `location_path` is still rewritten without touching `version` or
  `updated_at` — `test_repository_move_reindex_offline.py` still pins that — and optimistic
  concurrency is untouched.

## Follow-ups

- **A query in a script without word boundaries matches every item.** `normalize_search_text`
  drops everything NFKD cannot render as ASCII, so `q="電池"` normalizes to the empty string and
  `_item_matches_q` returns `True` for every item; an item named only in Japanese or Chinese is
  likewise unreachable by any query. Unchanged by this PR — the scan always answered it this way —
  and now written down in `test_a_query_that_normalizes_away_narrows_nothing`. Not filed: the
  shipped dictionaries are English and German, and a fix means changing what normalization means
  for every language at once, which is a decision rather than a bug fix. Worth an issue if the
  project takes a CJK language.
- `list_items` still materializes the whole filtered, sorted list before slicing a page. At 10,000
  items with `q` that is the 0.020 s above, so nothing to do now; noted because the scan is the
  only thing standing between a query and the item map.

## Handover

1. **Merged / left open** — this PR, for the M1 master to squash-merge once CI is green. Nothing
   left open for the owner.
2. **Test this by hand** — nothing that needs a phone, a German reader or the owner's own store.
   The behaviour change is a backend search result, pinned by tests.
3. **Validate locally** —
   1. `[dev-ha]` deploy this branch, create items named `Light` and `Flashlight`.
   2. `[browser]` search `light` in the card → **both** rows are listed. On `main` only `Light` is.
   3. `[browser]` search `caf` and `café` against an item named `Cafe` → both find it.
   4. `[dev-ha]` with ~2,000 seeded items: search, filter, sort, a subtree move and a location
      rename all answer as before, and search after the move follows the new display path.
   5. `[log]` nothing at ERROR carrying `haventory` through any of it.
4. **Decisions taken against drifted notes** — the issue comment counts the five fall-through
   tests as removable ("they pass trivially or become redundant"); the plan's PR 7 keeps them as
   the oracle, and that is what happened — four were kept and pass unchanged, and the fifth held
   nothing but assertions about the deleted pre-filter. The comment's "~249 LOC removed" measured
   `b878bfe`; the tree at `f1f6d9b` gave 258 from `repository.py`.
5. **Follow-ups** — above: the CJK normalization gap (not filed, with the reason) and the
   materialize-then-slice note (not filed). The behaviour fix earned #623, closed here.
6. **State left behind** — branch `claude/v0-8-0-m1-text-index`, two commits. No assets branch, no
   HA instance touched. Counting for this session: production removed 260 · tests removed 79 ·
   added 203.

---

### DIFF
--- a/custom_components/haventory/models.py
+++ b/custom_components/haventory/models.py
@@ -481,8 +481,9 @@ def normalize_search_text(text: str) -> str:
     """Normalize text for search matching (lowercase, strip accents).
 
     Uses NFKD normalization to separate accents from characters, then keeps only ASCII.
-    Shared by the repository search indexes and the post-filter in ``_item_matches_q``
-    so both layers agree on what "case-insensitive, accent-insensitive" means.
+    What ``_item_matches_q`` reads both the query and the item's text through, so
+    "case-insensitive, accent-insensitive" means one thing on both sides of the
+    comparison.
     """
 
     if not text:
--- a/custom_components/haventory/repository.py
+++ b/custom_components/haventory/repository.py
@@ -14,13 +14,12 @@ import base64
 import binascii
 import copy
 import json
-import re
 import uuid
 from collections import deque
 from collections.abc import Iterable, Iterator, Sequence
 from dataclasses import dataclass, replace
 from datetime import date
-from typing import Any, NamedTuple, TypedDict
+from typing import Any, TypedDict
 
 from .calendar_projection import next_occurrence_after
 from .exceptions import ConflictError, NotFoundError, ValidationError
@@ -58,7 +57,6 @@ from .models import (
     location_sort_key,
     monotonic_timestamp_after,
     new_uuid4,
-    normalize_search_text,
     normalize_text_for_sort,
     parse_uuid4,
     require_string_list,
@@ -76,21 +74,6 @@ from .models import (
 )
 
 LOGGER = context_logger(__name__)
-
-
-class _TextTokens(NamedTuple):
-    """Cached text-index tokens for one item, split by source.
-
-    ``base_*`` tokens derive from name/description/category/tags; ``path_*``
-    tokens derive from the denormalized ``location_path.display_path``. The
-    split lets subtree moves update only the path-derived buckets.
-    """
-
-    base_words: frozenset[str]
-    path_words: frozenset[str]
-    name_prefixes: frozenset[str]
-    base_trigrams: frozenset[str]
-    path_trigrams: frozenset[str]
 
 
 class PageResult(TypedDict):
@@ -116,9 +99,6 @@ class InternalIndexes(TypedDict):
 
 # Sentinel for optional args that distinguish "not provided" from explicit None
 UNSET: object = object()
-
-TRIGRAM_MIN_LEN = 3
-PREFIX_MIN_LEN = 2
 
 #: Longest pagination cursor this build will even attempt to decode. A cursor is
 #: base64 of a small JSON object this repository minted itself, so anything
@@ -252,14 +232,6 @@ class Repository:
         self._items_by_area_id: dict[str, set[str]] = {}
         # Cached name sort keys
         self._name_sort_key_by_item_id: dict[str, str] = {}
-
-        # Text Indices
-        self._word_to_item_ids: dict[str, set[str]] = {}
-        self._name_prefix_to_item_ids: dict[str, set[str]] = {}
-        self._trigram_to_item_ids: dict[str, set[str]] = {}
-        # Per-item cached tokens so clearing/delta-updating the text indexes
-        # never has to re-derive tokens from the item fields.
-        self._item_text_tokens: dict[str, _TextTokens] = {}
 
         # Location tree indexes
         self._children_ids_by_parent_id: dict[str | None, set[str]] = {}
@@ -460,9 +432,6 @@ class Repository:
         # Update subtree index
         self._add_item_to_subtree_index(item)
 
-        # Update text search index
-        self._index_item_text(item)
-
         # Increment generation on state modification
         self._increment_generation()
 
@@ -491,9 +460,6 @@ class Repository:
 
         # Remove from subtree index
         self._remove_item_from_subtree_index(item)
-
-        # Remove from text search index
-        self._clear_item_text_index(item)
 
         # Finally, drop from primary store
         self._items_by_id.pop(item_key, None)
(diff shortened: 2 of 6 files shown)


chrreiter:

| step | result |
| --- | --- |
| 1–2 `Light` / `Flashlight`, search `light` | **PASS** — the card lists both (`Flashlight`, `Light`), together with the two seeded items that also contain the query (`Ladekabel Lightning`, `USB-Kabel Sortiment`). |
| 3 `caf` / `café` find `Cafe` | **PASS** — both list exactly `Cafe`. |
| 4 search, filter, sort, subtree move, rename on a large store | **PASS** — 558 seeded items plus the 1,000 the `bulk` layer added and removed: `q` + `low_stock_only` answers the same rows as the 0.7.1 control; the Vorratsschrank subtree (49 items) renamed and moved away and back keeps every item's `version`/`updated_at` while every path follows, and the card's row shows the new path. |
| 5 `[log]` | **PASS** — nothing at ERROR carrying `haventory`; the sweep's only EXPECTED lines are the fuzz layers' contract WARNINGs. |

Evidence: [`m1/07-search-light.png`](https://raw.githubusercontent.com/chrreiter/HAventory/claude-assets-v0-8-0-l1/l1/m1/07-search-light.png). The full L1 record is on #236.

PR #626, opened 2026-08-23
refactor(repository): one chain walk, from_dict on the models, and the small copies

chrreiter:
## Summary

The last of §6.M1's repository work: BE-REPO-3, -4, -5, -6 and -8 from the
[2026-08-22 measurement on #230](https://github.com/chrreiter/HAventory/issues/230#issuecomment-5381712066).
Five copies of the same walk, one staged copy of two whole maps, two hand-written copies of the
model field lists, five counts that were one expression each, and three comments that told the
reader about the past instead of about the code.

- **One walk up the parent chain** (BE-REPO-3). `effective_area_id`, `_find_location_root`,
  `_get_ancestors`, the lineage `create_location` builds and the chain `_rebuild_paths_for_subtree`
  rebuilds each carried their own copy of the step ceiling and their own answer to a chain that
  closes into a loop. They now share `walk_location_chain`, which stops on an id it has already
  passed, and `location_chain_to_root`, which refuses a chain that reaches no root and names which
  of the two ways it failed. `build_location_path_from_map` — the helper two of the five were
  duplicating — is now written on those two, so the codebase holds one such loop.
- **A location update applied in place** (BE-REPO-4). `update_location` copied the whole location
  map and the whole children index, applied the change to the copies, rebuilt the subtree's paths
  against them and swapped them in. `_validate_parent_move` runs first, and after it the parent
  exists and is neither the node nor one of its descendants, so neither exception the rebuild can
  raise is reachable in a tree. `_rebuild_paths_for_subtree` loses the two optional map parameters
  that existed only for the staged copies.
- **`from_dict` beside `to_dict`** (BE-REPO-5), on `Item`, `Location` and `LocationPath`.
  `load_state` and `import_export._recompute_paths` stop building the models field by field, and
  the two `LocationPath` dict literals in the import path become `to_dict()` calls. The `sort_key`
  backfill lives once. `load_state` keeps its per-row `try`/`except`, so the `LoadReport` contract
  is untouched — `test_load_report_offline.py` is green and unedited.
- **One count, and the copies around the buckets** (BE-REPO-6). Five `_count_*` helpers → `_count`
  plus a named predicate per line; three places that inlined `_remove_from_bucket`'s
  discard-and-drop-empty dance now call it; the cached name sort key (one reader, which recomputed
  on a miss anyway) and the descendants map (filled and read inside one method) are gone;
  `_is_low_stock` was an alias for the models predicate.
- **Comments that state the constraint** (BE-REPO-8). The class docstring said a location change
  updates impacted items "to increment their version and `updated_at`" — the opposite of the
  invariant the rest of the codebase depends on. The "pre-WP4" pair collapses into one line on
  `LocationPath.from_dict`. The subtree-index removal narrated the call sequence that reaches it.
  `get_counts` no longer calls two live contract fields "legacy" or explains a past release.

No file left `custom_components/haventory/`, so `RETIRED_PATHS` is unchanged. No WS schema, error
code or payload field moved, so `docs/backend_api_contract.md` and `docs/data_shapes.md` are
unchanged.

Refs #230

## Counting

| | lines |
|---|---|
| Production removed | 544 |
| Test removed | 29 |
| Added | 491 |

From `git diff origin/main --numstat`: `repository.py` −485/+199, `models.py` −25/+194,
`import_export.py` −34/+10, tests −29/+88. Net −82; the production net is −141, and the difference
is the load block that moved out of `repository.py` into the two `from_dict`s.

## Type of change

## How was this tested?

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` → **1357 passed, 27 skipped**;
  formatted; `uv run mypy` → Success: no issues found in 26 source files.
  `npx vitest run` → **67 files, 1917 passed**, `npm run build` → 732.88 kB.
- [x] phacc (`scripts/test_integration.sh`, card built first): **153 passed in 16.51s** —
  `[ OK ] Integration tests OK`.

### Tests

Five tests were added and one was deleted.

- `test_every_stored_field_reads_back_into_the_model`,
  `test_a_row_that_carries_no_path_reads_as_the_empty_one` and
  `test_a_path_of_the_wrong_shape_is_refused` (`tests/test_serialized_shapes_offline.py`) —
  `from_dict` is the fourth place the field list is written out, so the file that holds the other
  three against each other holds the inverse too.
- `test_the_chain_walk_ends_on_a_loop_instead_of_following_it` and
  `test_a_chain_that_reaches_no_root_is_refused` (`tests/test_models_offline.py`) — the shared
  walk's two stop conditions, which is the part of BE-REPO-3 the issue called medium risk.
- `test_update_location_is_atomic_when_path_rebuild_fails`
  (`tests/test_repository_locations_offline.py`) is **deleted, not rewritten**, per the subtraction
  rule. It monkeypatched `_rebuild_paths_for_subtree` into raising `RuntimeError`, which is the
  failure the staged maps existed for; with the staging gone there is nothing left for it to
  assert that validation does not already exclude.

The autouse index-invariant oracle (`tests/repository_invariants.py`, PR 8) runs after every test
over every repository built, so it covered each of these commits as they landed.

The last commit is a correction found by diffing the load path against `main` rather than by a
failing test: the first cut of `LocationPath.from_dict` read *anything* that was not an object as
the empty path, so a stored `location_path` of `"garbage"` loaded with no path instead of being
dropped and named in the `LoadReport`. Absent is old; present and the wrong shape is corrupt. The
third new test above is the pin.

## Checklist

- [x] Docs updated where behavior changed — no behaviour changed; see "Differences" below.
- [x] WebSocket API changes keep `ws.py`, `docs/backend_api_contract.md`, and `docs/data_shapes.md`
      in sync — none in this PR.
      optimistic `version`).

## Differences a reader should check rather than assume

Four places where the collapse could not be byte-identical. None is reachable from a store or a
document this integration wrote itself.

1. **`_get_ancestors` on a chain that loops.** The old walk stopped *before* recording the id that
   closes the loop; the shared walk stops *after*, so the returned list can carry that id (and
   therefore a repeat). Its two callers add to and remove from the same buckets, and both are
   idempotent, so the index is identical either way — and a cyclic location makes setup refuse to
   load at all.
2. **`create_location` with a broken ancestor above a parent that exists.** The message changes
   from "parent_id must reference an existing location" to "location_id must reference an existing
   location chain" / "location graph too deep or cyclic", which is what every other chain builder
   already answers.
3. **`build_location_path_from_map` on a cyclic map** now answers after the cycle rather than after
   10,000 steps. Same message, same refusal.
4. **A location name in an import document** is now trimmed, because `Location.from_dict` runs it
   through `validate_required_name` — which is what `load_state` already did to the same field, and
   what the interactive write path does. Only a hand-authored document with padded names differs.

`get_counts` also reads `today_local_date()` once instead of five times, so a call that straddles
midnight now answers about one day rather than two.

None — no card file is touched.

## Handover

1. **Merged / left open** — this PR, left open for the M1 master to review and squash-merge; it is
   the last of §6.M1's nine.
2. **Test this by hand** — `[owner]` nothing specific. The one path no test reaches is a store the
   owner's own install has grown: loading it should produce the same counts and the same location
   tree as before, which the diagnostics panel shows at a glance.
3. **Validate locally**
   1. `[dev-ha]` deploy this branch and reload the entry from Settings → Devices & services.
      Expected: the entry reloads, the item and location counts are what they were.
   2. `[browser]` rename a location that has a subtree and items in it. Expected: every descendant
      path and every item's location line update, and the items keep their positions in a
      "recently updated" sort — the rename must not restamp them.
   3. `[browser]` move a subtree under another node, then move it back. Expected: paths follow both
      ways; no item's version changes (visible as no concurrency error on a subsequent edit).
   4. `[browser]` set an area on a location deep in a tree. Expected: the area lands on the tree's
      root and every item under it reports it.
   5. `[browser]` export the inventory, then import the document back. Expected: valid, and the
      report classifies every entity as unchanged.
   6. `[log]` nothing at ERROR carrying `haventory` through all of the above.
4. **Decisions taken against drifted notes**
   - BE-REPO-6(e) lists three one-line wrappers. Only `_is_low_stock` went. `_low_stock_group`
     encodes the 0/1 block convention `_tuple_cmp` compares on, and inlining it at its two call
     sites would need that convention spelled out twice. `_reindex_item_replacement` has three
     callers and a required order — unindexing ends by dropping the id from the primary store, so
     indexing first would leave the repository without the item — which is now its docstring
     instead of the comment that restated the two calls below it.
   - BE-REPO-3's suggested `_chain(loc_key, loc_map) -> list[Location]` became a generator in
     `models.py` plus a raising list builder beside it. `models.py` owns
     `LOCATION_GUARD_MAX_STEPS`, `Location` and the path builders, and putting the walk there is
     what let `build_location_path_from_map` — the sixth copy of the loop, which the issue notes
     two of the five were duplicating — be written on it rather than left standing.
   - BE-REPO-8 also lists the `get_counts` "legacy" sentence; it is fixed here because the
     surrounding method was being rewritten for BE-REPO-6(a) anyway. The `#225` forward pointer at
     `LoadReport` stays, as the issue asks.
5. **Follow-ups** — `tests/test_repository_roundtrip_offline.py`'s
   `test_legacy_store_without_sort_key_is_backfilled_on_load` still says "Pre-WP4 stores" in its
   docstring. Not filed: it is a one-line wording fix in a file this PR does not otherwise touch,
   and M6's docs pass is where it belongs.
6. **State left behind** — branch `claude/v0-8-0-m1-repo-copies`, six commits, no assets branch,
   no HA instance touched. Counting for this PR: 544 production lines removed, 29 test lines
   removed, 491 added.

### DIFF
--- a/custom_components/haventory/import_export.py
+++ b/custom_components/haventory/import_export.py
@@ -649,19 +649,12 @@ def _recompute_paths(
     ``None`` when a reference is broken (with ``errors`` populated).
     """
 
-    loc_objs: dict[str, Location] = {}
-    for lid, d in locations.items():
-        loc_objs[lid] = Location(
-            id=parse_uuid4(str(d["id"]), field_name="location.id"),
-            parent_id=(
-                parse_uuid4(str(d["parent_id"]), field_name="location.parent_id")
-                if d.get("parent_id") is not None
-                else None
-            ),
-            name=str(d["name"]),
-            area_id=str(d["area_id"]) if d.get("area_id") is not None else None,
-            path=EMPTY_LOCATION_PATH,
-        )
+    # The path a document carries is dropped rather than read: it is recomputed
+    # below from the names and parent links this import is about to store, and a
+    # hand-edited one would otherwise decide whether the document loads at all.
+    loc_objs: dict[str, Location] = {
+        lid: Location.from_dict({**d, "path": None}) for lid, d in locations.items()
+    }
 
     # Recompute location paths (also validates parent references resolve).
     out_locations: dict[str, Any] = {}
@@ -671,18 +664,7 @@ def _recompute_paths(
         except ValidationError as exc:
             errors.append(_err(f"locations[{lid}].parent_id", str(exc)))
             return None
-        out_locations[lid] = {
-            "id": str(obj.id),
-            "name": obj.name,
-            "parent_id": str(obj.parent_id) if obj.parent_id is not None else None,
-            "area_id": obj.area_id,
-            "path": {
-                "id_path": [str(x) for x in path.id_path],
-                "name_path": list(path.name_path),
-                "display_path": path.display_path,
-                "sort_key": path.sort_key,
-            },
-        }
+        out_locations[lid] = {**obj.to_dict(), "path": path.to_dict()}
 
     out_items: dict[str, Any] = {}
     for iid, d in items.items():
@@ -693,17 +675,11 @@ def _recompute_paths(
             )
             return None
         if loc_id is not None:
-            lp = build_location_path_from_map(
+            location_path = build_location_path_from_map(
                 parse_uuid4(str(loc_id), field_name="item.location_id"), locations_by_id=loc_objs
-            )
-            location_path = {
-                "id_path": [str(x) for x in lp.id_path],
-                "name_path": list(lp.name_path),
-                "display_path": lp.display_path,
-                "sort_key": lp.sort_key,
-            }
+            ).to_dict()
         else:
-            location_path = {"id_path": [], "name_path": [], "display_path": "", "sort_key": ""}
+            location_path = EMPTY_LOCATION_PATH.to_dict()
         entry = dict(d)
         entry["tags"] = normalize_tags(d.get("tags") or [])
         entry["custom_fields"] = dict(d.get("custom_fields") or {})
--- a/custom_components/haventory/models.py
+++ b/custom_components/haventory/models.py
@@ -94,6 +94,32 @@ class LocationPath:
             "sort_key": self.sort_key,
         }
 
+    @classmethod
+    def from_dict(cls, data: object) -> LocationPath:
+        """Read back what ``to_dict`` wrote, tolerating an absent nesting.
+
+        ``sort_key`` is backfilled from ``display_path`` when the payload
+        carries none: stores written before it was persisted have to sort
+        alongside the ones that were, and it is derived, so recomputing it
+        costs nothing but the read.
+
+        An entry of ``id_path`` that is not a UUID v4 raises: the id path is
+        what a subtree filter matches on, so a row carrying a broken one is
+        corrupt rather than merely old.
+        """
+
+        raw = data if isinstance(data, Mapping) else {}
+        display = str(raw.get("display_path", ""))
+        return cls(
+            id_path=[
+                parse_uuid4(str(entry), field_name="path.id_path")
+                for entry in (raw.get("id_path") or [])
+            ],
+            name_path=list(raw.get("name_path") or []),
+            display_path=display,
+            sort_key=str(raw.get("sort_key", "")) or normalize_text_for_sort(display),
+        )
+
 
 EMPTY_LOCATION_PATH = LocationPath(id_path=[], name_path=[], display_path="", sort_key="")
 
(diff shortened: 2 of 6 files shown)


chrreiter:

| step | result |
| --- | --- |
| 1 reload the entry; counts unchanged | **PASS** — 558 / 92 before and after. |
| 2 rename a location with a subtree and items | **PASS** — `Vorratsschrank` (49 items over three shelves) → `Vorratsschrank L1`: every descendant path and every item's location line updated; every item kept its `version` and `updated_at` (the "recently updated" order did not move). |
| 3 move the subtree under another node and back | **PASS** — under `Garage` and back under `Küche`: paths follow both ways, no item's version changed. |
| 4 set an area on a deep location | **PASS** — `area_id: bedroom` on `Küche / Vorratsschrank / Oberstes Fach` lands on the tree's root (`Küche.area_id = bedroom`, the deep node stays `null`) and an item under it reports `effective_area_id: bedroom`; put back to `kuche`. |
| 5 export, then import the document back | **PASS** — valid; every one of 558 items and 92 locations classified `unchanged`. |
| 6 `[log]` | **PASS** — nothing at ERROR through all of it. |

PR #627, opened 2026-08-24
refactor(models): every field rule written once

chrreiter:
## Summary

An item's fields were validated twice: once inline in `create_item_from_create`, once in one of the nine `_update_*` helpers. The two copies had drifted into different shapes for the same rule — the name cap written three times, the threshold check twice, the growth caps reading `previous` in create and update by different routes.

They are now one ordered table of per-field rules that a create and an update walk alike:

- `_ItemWrite` carries the draft the rules fill in — the field defaults on a create, a copy of the stored item on an update. Every cap that refuses *growth* reads what the item already carries off that same draft, so on a create it is an absolute cap and no second shape is needed.
- The nine `_update_*` helpers and the inline create body become twelve rules in `_ITEM_FIELD_RULES`, one per field or per bound pair (`checked_out`/`due_date`, the reminder's date and interval, `location_id` and its denormalized path).
- `validate_location_name` becomes `validate_write_name`: trim plus `NAME_MAX_LENGTH` is the write path's name rule for an item and a location alike, and it was written three times.
- `validate_item_status` and `coerce_item_status` were the same four lines twice, differing in the last one. One function with `default=`: a write path passes none and gets the refusal, a load path passes the status an unreadable stored value should read as.
- `new_uuid4_str` had zero production callers; the fifteen test call sites take `str(uuid.uuid4())`.

No message text changed and the update path answers byte-identically. Verified against the pre-change implementation with a throwaway differential harness (both implementations imported side by side, outcomes compared field by field): every single-field create, every single-field update over five base items with and without a location map, the grandfathered pre-cap cases, and the reminder-anchor paths all match. The one difference the harness found is noted under Follow-ups.

Refs #230 (BE-MODELS-C3), §6.M2 PR 1.

## Counting

| | lines |
|---|---:|
| production removed | 238 |
| test removed | 30 |
| added | 253 |

`git diff --stat` against `origin/main`. The net is small because each rule now carries in one place the constraint the two inline copies encoded between them: the executable statements in the create/update/status region go from 134 to 107 (`ast`, docstrings excluded), and `models.py` from 2144 to 2130 lines.

## Type of change

## How was this tested?

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` (1357 passed, 27 skipped), `uv run ruff check .`, `uv run ruff format --check .`, `uv run mypy` — all clean
- [x] Frontend (`cards/haventory-card`): `npx eslint .`, `npm run typecheck`, `npx vitest run` (67 files, 1917 tests passed), `npm run build`
- [x] phacc (`scripts/test_integration.sh`, card built first): **153 passed in 16.88s**

No test was rewritten to pass. The test edits are the twin's call sites moving onto the merged function, two test names that said "coerce", and `new_uuid4_str()` becoming `str(uuid.uuid4())`.

## Checklist

- [x] PR title follows Conventional Commits
- [x] Tests added/updated — a cut changes no behaviour, so the existing suite is the pin (the sixteen date-message cases from #590, `test_item_input_validation_offline.py`'s cap and grandfathering cases, `test_item_status_offline.py`)
- [x] Docs updated where behavior changed — none: no WS schema, error code or field moved, so `docs/backend_api_contract.md` and `docs/data_shapes.md` are untouched
- [x] WebSocket API changes keep `ws.py` and the two contract docs in sync — no API change
- [x] Essential invariants preserved (case-insensitive search, denormalized `location_path` rebuilt on every write, optimistic `version`)
- [x] No file removed from `custom_components/haventory/`, so `RETIRED_PATHS` is unchanged

- **The second `CUSTOM_FIELDS_MAX_KEYS` check stays.** The #230 comment (§2.1) reads it as re-applying what `validate_custom_fields` already did. It does not: the check inside `validate_custom_fields` bounds the *patch*, this one bounds the *merged map*. A two-key patch onto an item at 49 keys passes the first and is caught by the second, which is what `test_an_item_over_the_custom_field_cap_may_not_grow_further` pins. Dropping it would let an item walk past the cap one patch at a time. It is now written once, in the one rule that owns the field.
- **`validate_optional_date(value, *, field_name)` and the `field_name` on `normalize_date_yyyy_mm_dd` were already in the tree** (#590), so the twin date validators and the "due_date must be…" wording defect that C3 also listed are not part of this PR.

## Follow-ups

- A create payload that is wrong about **two** fields at once can now name the other one of them in its single refusal. The two paths validated in different orders (create checked `location_id` last, an update checks it before tags and category), and one table has one order — the update path's, which is byte-identical. Every message text is unchanged and each refusal is true of the payload, so this is not filed: an API contract names codes and texts, not which of two invalid fields wins the race.

## Handover

1. **Merged / left open** — this PR, for the M2 master to merge once CI is green.
2. **Test this by hand** — nothing. No user-visible surface, no dictionary key, no stored shape moves.
3. **Validate locally** — nothing: pinned by `tests/test_item_input_validation_offline.py`, `tests/test_models_offline.py`, `tests/test_item_status_offline.py` and the phacc run above.
4. **Decisions taken against drifted notes** — the two above.
5. **Follow-ups** — the create multi-error ordering note above; not filed.
6. **State left behind** — branch `claude/v0-8-0-m2-field-rules`, no assets branch, no HA instance touched. Counting: production removed 238 · tests removed 30 · added 253.

### DIFF
--- a/custom_components/haventory/models.py
+++ b/custom_components/haventory/models.py
@@ -17,9 +17,10 @@ from __future__ import annotations
 import re
 import unicodedata
 import uuid
-from collections.abc import Collection, Iterable, Iterator, Mapping, Sequence
+from collections.abc import Callable, Collection, Iterable, Iterator, Mapping, Sequence
 from dataclasses import dataclass, field, replace
 from datetime import UTC, datetime, timedelta
+from functools import partial
 from typing import Any, Final, Literal, NotRequired, TypedDict
 
 from homeassistant.util import dt as dt_util
@@ -661,11 +662,11 @@ def validate_required_name(value: object) -> str:
     return value.strip()
 
 
-def validate_location_name(name: str) -> str:
-    """Validate a location name and return a trimmed value.
+def validate_write_name(name: object) -> str:
+    """Validate a name a client wrote and return a trimmed value.
 
-    The write path's rule: what the load path requires, plus the length cap the
-    load path deliberately does not apply.
+    The write path's rule, for an item and a location alike: what the load path
+    requires, plus the length cap the load path deliberately does not apply.
     """
 
     trimmed = validate_required_name(name)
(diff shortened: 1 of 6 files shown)


chrreiter:

| step | result |
| --- | --- |

PR #628, opened 2026-08-24
refactor(models): filter_items as a predicate list

chrreiter:
## Summary

`filter_items` decided sixteen things about every item, whichever ones the filter actually asked
for. Fourteen `bool(flt.get("x")) if "x" in flt else False` reads, a seventeen-line disjunction
whose only job was to notice that none of them was on, and then sixteen booleans computed eagerly
per item and `and`ed afterwards — `q`, the only test that normalizes text, among them.

The keys a filter carries now become a list of predicates, built once per query by
`_filter_predicates`. An item is walked against those and no others, and stops at the first one it
fails. `q` is appended last, so a cheap flag drops what it can before any text is normalized.

Also here:

- The five `item_*_is_*` predicates are the five `(field, inclusive)` pairs of one `_date_passed`.
  The five public names stay as one-liners, and `today` stays a parameter the caller fills once per
  query from `today_local_date()` — `models.py` still reads no `hass`.
- Two small tables replace two hand-written runs: `_DATE_FILTERS` (five keys) and
  `_TIMESTAMP_BOUNDS` (four bounds, in the order a filter carrying more than one malformed bound is
  refused in).
- `sort_items` no longer re-checks the field and order that `validate_sort` already refused at the
  WebSocket boundary.
- `_item_matches_locations` loses its `if not needles: return True` guard: the predicate is built
  only when the filter names locations, so an empty parsed selection now selects nothing by
  reading its own loop rather than through a second early return in `filter_items`.

Refs #230 (BE-MODELS-C4, and item 1's scan — `_item_matches_q` applied by `filter_items` is still
the declared contract).

### Behaviour

None. Every value the old function read, it reads in the same order, so a filter carrying more than
one malformed value is refused by the same one; a location selection that parses to no id still
matches nothing; `updated_before: ""` still keeps nothing and `updated_after: ""` still keeps
everything; a filter carrying only presentation hints (`low_stock_first`) still returns the whole
population.

The one caller-visible edge is `sort_items` called directly with a field or order outside the
vocabulary: it now orders by the default instead of raising. No production caller can reach that —
`ws.py` calls `validate_sort` first, and `import_export.py` and `calendar.py` pass no sort at all.

### Benchmark (`ASSERT_BUDGETS=1`, 10,000 items, best of 5 passes)

`tests/test_repository_search_benchmark_offline.py`, run back to back on the same host:

| | before this PR | after this PR |
|---|---:|---:|
| `q` list through `repo.list_items` | 0.013 s | 0.012–0.014 s |

Unchanged, as expected: with `q` alone there is nothing to short-circuit *to*, and the fifteen other
booleans the old loop computed per item are cheap next to `normalize_search_text`. Where the
predicate order pays is a query that carries a flag as well, measured against `filter_items`
directly over the same 10,000 items so the repository's index pre-filter does not narrow the
population first:

| query | before | after |
|---|---:|---:|
| `{"q": "Widget"}` | 0.0127 s | 0.0128 s |
| `{"q": "Widget", "category": "Category 3", "checked_out": False}` | 0.0139 s | 0.0046 s |
| `{"category": "Category 3"}` | 0.0035 s | 0.0035 s |

Adding a flag to a `q` list used to make it strictly more work; it now makes it three times less.
Budget (0.5 s) is untouched, and item 1's 0.012 s figure still holds.

### Counting

| | lines |
|---|---:|
| production removed | 154 |
| test removed | 22 |
| added | 179 |

From `git diff --stat` against `origin/main`: `models.py` 179 insertions / 154 deletions,
`tests/test_models_filter_sort_offline.py` 0 / 22. The insertions are the predicate builder, the two
factories, the two tables and their notes; `filter_items` itself goes from 154 lines to 15.

`repository.py`'s five `_count_*` are named by the plan as part of this cut and were already
collapsed by #626 — they are `self._count(population, predicate)` calls over the same five public
predicates, so they ride `_date_passed` from this PR without an edit. Nothing left to remove there.

## Type of change

## How was this tested?

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` → **1355 passed, 27 skipped**;
      formatted; `uv run mypy` → Success: no issues found in 26 source files
      `npx vitest run` → **67 files, 1917 tests passed**, `npm run build` → built in 228 ms
- [x] phacc (`scripts/test_integration.sh`, card built first): **`153 passed in 14.48s` — `[ OK ]
      Integration tests OK`**

`tests/test_models_filter_sort_offline.py` is the pin for the semantics and passes unchanged apart
from the two deletions below — that whole file, and the WebSocket item/list tests, went green on the
first run of the rewritten function.

### The two deleted tests

Both pinned only the duplicate re-check inside `sort_items`, and the refusal a caller actually sees
stays pinned where it is answered:

- `test_sort_invalid_inputs_raise` — a bad field and a bad order. Both are asserted against
  `validate_sort` by `test_validate_sort_accepts_the_vocabulary_and_rejects_the_rest` (`sort.field`
  for `colour`, `sort.order` for `sideways`), and on the wire by
  `tests/test_ws_items_offline.py::test_item_list_refuses_an_unknown_sort_field`.
- `test_validate_sort_agrees_with_sort_items` — its loop over `SORT_FIELDS` sorted an **empty**
  list, which `sort_items` returns before it reads `field` at all, so that half asserted nothing;
  the rest was the same duplicate raise. It never proved every `SORT_FIELDS` member has a branch —
  the chain's `else` has always absorbed one that does not, before this PR as after it.

## Checklist

- [x] Tests added/updated — a cut: no test rewritten to pass, two deleted with the duplicate they
      pinned, the rest left alone.
- [x] Docs updated where behavior changed — none: no schema, error code or field moves, so
      `docs/backend_api_contract.md` and `docs/data_shapes.md` are unchanged.
- [x] WebSocket API changes keep `ws.py`, `docs/backend_api_contract.md` and `docs/data_shapes.md`
      in sync — no WS change.
- [x] Essential invariants preserved (case-insensitive search through the same `_item_matches_q`,
      denormalized `location_path` read by the same subtree test, optimistic `version` untouched).

None — no card change.

## Follow-ups

- The `q` + flag short-circuit is not pinned by a benchmark case. A committed one would want a
  narrowing key the repository does **not** index, and category, status, `tags_any`,
  `checked_out: true`, `low_stock_only` and location are all indexed — so the case that shows the
  gain is not the case a household runs. Named, not filed: a performance property, not a
  correctness one.
- `subscriptions._payload_inspection_is_overdue` is still the third copy of
  `item_inspection_is_overdue`, over the serialized dict rather than the `Item`. Its docstring says
  why it cannot call the shared one, and the plan puts it elsewhere. Not filed.

## Handover

1. **Merged / left open** — this PR, for the master to merge once CI is green. Nothing stacks on it
   inside M2 except PR 3, which touches the same file.
2. **Test this by hand** — nothing. No user-facing string, no lifecycle, no card change.
3. **Validate locally** — nothing new: the filter and sort semantics are pinned by
   `tests/test_models_filter_sort_offline.py` (unchanged apart from the two deletions above) and by
   the WebSocket `item/list` tests. If the master is driving its own HA anyway: `[dev-ha]` an
   `item/list` with `filter: {"q": "…", "low_stock_only": true}` and one with
   `sort: {"field": "colour", "order": "asc"}` — the first answers as before, the second still comes
   back `validation_error`.
4. **Decisions taken against drifted notes** — the plan's "`repository.py`'s `_count_*` collapse
   onto the same predicates" was already done by #626; they ride the new `_date_passed` unedited.
   The plan offered "drop the re-validation in `sort_items` **or** drop `validate_sort`"; §6.M2 PR 2
   picks the former, which is what this does.
5. **Follow-ups** — the two above, neither filed.
6. **State left behind** — branch `claude/v0-8-0-m2-predicate-list`; no assets branch; no HA
   instance touched. Counting for this PR: production removed 154 · tests removed 22 · added 179.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YN6wQcqZD9UGhcDjXmVb8L)_

### DIFF
--- a/custom_components/haventory/models.py
+++ b/custom_components/haventory/models.py
@@ -1770,19 +1770,32 @@ def today_local_date() -> str:
     return dt_util.now().date().isoformat()
 
 
+def _date_passed(item: Item, field: str, today: str, *, inclusive: bool) -> bool:
+    """Whether the item's ``field`` date has come round by ``today``.
+
+    The one test behind the five named below, which are its five (field,
+    inclusive) pairs. An item carrying no date on that field is never counted.
+    Both sides are YYYY-MM-DD, which compares correctly as text. An empty
+    ``today`` reads the instance's current local date, and reads it only for an
+    item that has a date to compare — callers walking many items fill it in
+    once rather than paying for the clock per item.
+    """
+
+    value: str | None = getattr(item, field)
+    if not value:
+        return False
+    day = today or today_local_date()
+    return value <= day if inclusive else value < day
+
+
 def item_is_overdue(item: Item, *, today: str = "") -> bool:
     """Return True when the item's due date has passed.
 
     A due date only exists while an item is checked out (see
     ``validate_due_date_rules``), so this needs no separate checked-out test.
-    Both sides are YYYY-MM-DD, which compares correctly as text. ``today``
-    defaults to the instance's current local date; callers filtering many items
-    pass it in once rather than re-reading the clock per item.
     """
 
-    if not item.due_date:
-        return False
-    return item.due_date < (today or today_local_date())
+    return _date_passed(item, "due_date", today, inclusive=False)
 
 
 def item_is_due(item: Item, *, today: str = "") -> bool:
@@ -1795,22 +1808,17 @@ def item_is_due(item: Item, *, today: str = "") -> bool:
     ``item_inspection_is_due`` has to ``item_inspection_is_overdue``.
     """
 
-    if not item.due_date:
-        return False
-    return item.due_date <= (today or today_local_date())
+    return _date_passed(item, "due_date", today, inclusive=True)
 
 
 def item_inspection_is_overdue(item: Item, *, today: str = "") -> bool:
     """Return True when the item is past the date it was next due for inspection.
 
     Independent of the check-out state: an inspection is a fact about the item,
-    not about a borrowing, so this walks any item that carries a date. Same
-    text comparison and same ``today`` convention as ``item_is_overdue``.
+    not about a borrowing, so this walks any item that carries a date.
     """
 
-    if not item.inspection_date:
-        return False
-    return item.inspection_date < (today or today_local_date())
+    return _date_passed(item, "inspection_date", today, inclusive=False)
 
 
 def item_inspection_is_due(item: Item, *, today: str = "") -> bool:
@@ -1823,9 +1831,7 @@ def item_inspection_is_due(item: Item, *, today: str = "") -> bool:
     items whose date is today, and every overdue item is also due.
     """
 
-    if not item.inspection_date:
-        return False
-    return item.inspection_date <= (today or today_local_date())
+    return _date_passed(item, "inspection_date", today, inclusive=True)
 
 
 def item_reminder_is_due(item: Item, *, today: str = "") -> bool:
@@ -1837,9 +1843,7 @@ def item_reminder_is_due(item: Item, *, today: str = "") -> bool:
     date the household said to be reminded on.
     """
 
-    if not item.reminder_date:
-        return False
-    return item.reminder_date <= (today or today_local_date())
+    return _date_passed(item, "reminder_date", today, inclusive=True)
 
 
 def _parse_location_selection(location_ids: Sequence[str]) -> list[uuid.UUID]:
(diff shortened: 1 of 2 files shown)


chrreiter:

| step | result |
| --- | --- |

PR #629, opened 2026-08-24
refactor(init): the manifest is already a constant; the /local path never shipped; the unreachable checks

chrreiter:
## Summary

Four cuts in the lifecycle modules, none of them user-visible. The ids are the ones the
[2026-08-22 measurement comment on #230](https://github.com/chrreiter/HAventory/issues/230) uses.

**L4 — the card URL's version is already a constant.** `_async_card_url` asked HA's loader for
the manifest and fell back to an executor file read, so `_read_manifest_version`,
`_async_manifest_version` and `_MANIFEST_PATH` existed to reach a string `INTEGRATION_VERSION`
already holds: release-please rewrites the two together and
`tests/test_release_version_consistency.py` fails when they disagree, and `ws.py` has served the
constant as `haventory/version` all along. The URL is byte-identical, so `_card_url` is sync and
its three call sites stop awaiting it; its fallback to the bare path went with the readers,
since a constant is always there.

**L5 — the `/local` path never shipped.** `_LEGACY_CARD_URL_PATH` matched a Lovelace resource
pointing at `/local/haventory/haventory-card.js`. The bundle moved into the integration package
in `ea8d038`, which `git tag --contains` reports as v0.1.0, so no tagged release ever served that
path and only an untagged pre-0.1.0 checkout could hold such a resource. `_points_at_card` now
compares against the one served path.

**L7a — the checks the load path already makes, and one normalizer.**
`_validate_storage_payload` ran *after* `DomainStore.async_load`, which either returns
`_empty_payload()`, raises for a store a newer schema wrote, or stamps its own version — so the
payload it saw always carried this build's version and its two version branches could not fire
from setup. What they refuse is refused inside `DomainStore` and pinned there by
`tests/test_storage_offline.py`. The one reachable check — `items` and `locations` are maps of
rows, which a hand-edited file need not hold — moves into `async_load` beside the normalisation
it guards, and setup still answers it with `ConfigEntryNotReady` through the `StorageError` it
already maps. The `STORE_COLLECTIONS` defaults were written out at four sites in two spellings;
`_normalized(payload, schema_version=…)` is the one they share, with the payload still winning
over the defaults everywhere, extra keys included. `schema_downgrade_message` had one caller left
and is inlined into it word for word.

**L7b — one raw copy before a lossy load, not two.** `_async_settle_lossy_load` copied the store
aside under the same key `repairs.py` had just copied it to, and its own docstring said the
second write was the same bytes. A load reaches settle only with `CONF_ALLOW_LOSSY_LOAD` set,
that option is set in exactly one place, and the flow that sets it withdraws the offer when its
own copy fails — so the copy is already on disk, holding the rows the rewrite drops. The refusal
branch goes with it: the one that can still refuse is the flow's.

Refs #230

## Counting

| | lines |
|---|---|
| production removed | 178 (`__init__.py` 121, `storage.py` 57) |
| test lines removed | 346 (`test_frontend_registration.py` 189, `test_init_offline.py` 93, `conftest.py` 48, `test_entry_removal_offline.py` 16) |
| lines added | 146 (production 84, tests 59, skill harness 3) |

`git diff --shortstat origin/main...HEAD`: 8 files changed, 146 insertions(+), 532 deletions(-).

## Type of change

## How was this tested?

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` (1338 passed, 27 skipped), `uv run ruff check .`, `uv run ruff format --check .`, `uv run mypy` (26 source files, no issues)
- [x] Frontend (`cards/haventory-card`): `npx eslint .`, `npm run typecheck`, `npx vitest run` (67 files, 1917 tests passed), `npm run build`
- [x] phacc, card built first — `scripts/test_integration.sh`: **`153 passed in 14.82s`**

`tests/integration/test_repairs.py` is the gate for the backup cut, and it is in that run:
`test_a_corrupt_row_offers_a_fix_that_backs_up_reloads_and_clears` still finds the dropped row in
`haventory_store_corrupt_backup` after the flow, and
`test_the_repair_survives_a_restart_with_no_mutation_in_between` still finds the store rewritten
without it and no refusal on the restart.

### What moved on the test side

- The four tests that drove the loader and the executor read go, and
  `test_registered_url_carries_the_manifest_version` stays as
  `test_registered_url_carries_the_integration_version`, comparing the registered URL against the
  constant. `test_migrates_a_legacy_local_resource_in_place` goes with the `/local` path.
- `test_collapses_duplicate_card_resources` keeps testing the collapse, with a stale `?v=` entry
  as the duplicate instead of the retired legacy URL — the self-healing it pins is untouched.
- `test_setup_entry_invalid_version_raises` and the two direct `_validate_storage_payload` tests
  go: the first monkeypatched `async_load` into returning a payload that method cannot return,
  and the other two re-tested what `test_storage_offline.py` pins on `DomainStore`.
  `test_setup_entry_invalid_collections_raise` becomes
  `test_setup_entry_retries_on_a_collection_of_the_wrong_type` and seeds a real store instead of
  patching the load.
- The stub's `async_get_integration` and its `hass.executor_jobs` recorder lose their last
  readers and go; the stub is that much closer to real HA, which records neither.
- `.claude/skills/run-haventory`'s lifecycle probe loses its two `/local` variants: it would
  otherwise assert an entry gets rewritten that the integration now leaves alone.

## Checklist

- [x] Tests added/updated — a cut's tests are deleted or moved, never rewritten to pass; the two rewritten ones (the collections check, the duplicate collapse) test the same production behaviour through a path that still exists.
- [x] Docs updated where behavior changed — none did; no doc names the manifest read, the `/local` path or the second copy.
- [x] WebSocket API changes keep `ws.py`, `docs/backend_api_contract.md`, and `docs/data_shapes.md` in sync — no WS schema, error code or field moved.
- [x] Essential invariants preserved (case-insensitive search, denormalized `location_path`, optimistic `version`).

## Follow-ups

- `async_persist_immediate` (`storage.py`) is now a one-line wrapper around `async_persist_repo`
  whose only reason to exist is its DEBUG line; its two callers are the settle rewrite and the
  teardown flush. Not filed: it is a comment-sized finding, and M1's debounce cut already looked
  at this function.
- `DomainStore.async_save` writes `schema_version` first now rather than appending it after the
  caller's keys, so the key order inside `.storage/haventory_store` changes on the next write.
  Nothing reads the file by key order. Not filed.

## Handover

1. **Merged / left open** — this PR is for the M2 master to merge once CI is green; nothing in
   the package stacks on it.
2. **Test this by hand** — nothing `[owner]`-only. The card is byte-identical and no string
   changed, so nothing `[German]` or `[phone]`.
3. **Validate locally**
   1. `[dev-ha]` deploy the branch and restart; over WS, `lovelace/resources` shows exactly one
      HAventory entry and its URL ends `?v=<INTEGRATION_VERSION>` (`0.7.1` on this branch) — the
      same string the sidebar panel's `module_url` carries.
   2. `[browser]` open a dashboard with the card and `/haventory`: both load, and the browser's
      network tab shows one request for `haventory-card.js`, not two.
   3. `[log]` grep the start-up log for a loop-protection stack trace naming `manifest.json` —
      there is none to find now, and there was none before either; the point is that the read is
      gone rather than merely offloaded.
   4. `[dev-ha]` optional, only if the store is worth breaking: hand-edit an item row into
      something unreadable, restart, press **Fix** in Repairs, restart again — the entry comes up
      both times and `haventory_store_corrupt_backup` holds the broken row.
4. **Decisions taken against drifted notes** — the measurement comment sizes L7's dict-type check
   as "moves into `async_load` as ~4 lines"; it landed as a loop over `("items", "locations")`
   with its own message per collection, since one message naming both read as if both were
   wrong. It also lists five `STORE_COLLECTIONS` default sites while the plan text says four —
   all five are now `_normalized`.
5. **Follow-ups** — the two above, neither filed.
6. **State left behind** — branch `claude/v0-8-0-m2-init-leftovers`, no assets branch, no HA
   instance touched. Counting for this PR: 178 production lines removed, 346 test lines removed,
   146 added.

### DIFF
--- a/.claude/skills/run-haventory/SKILL.md
+++ b/.claude/skills/run-haventory/SKILL.md
@@ -626,10 +626,8 @@ uv run python .claude/skills/run-haventory/lifecycle_probe.py all --yes
 ```
 
 `resources` sets the Lovelace resource to each shape a restart can find — hand-pinned
-`?v=<hash>`, stale `?v=<old version>`, bare URL, plus the two legacy
-`/local/haventory/haventory-card.js` shapes an install predating the integration-served
-bundle carries — restarts, and asserts one entry survives **under the original resource
-id**; a second entry would load the card module twice and the second
+`?v=<hash>`, stale `?v=<old version>`, bare URL, a URL carrying a second query parameter
+— restarts, and asserts one entry survives **under the original resource id**; a second entry would load the card module twice and the second
 `customElements.define` would throw. `downgrade` writes a higher `schema_version`
 into the store and asserts the entry lands in `setup_error` (not `setup_retry` — retrying
 cannot teach this build a newer schema) with the payload untouched. `entry` removes the
--- a/.claude/skills/run-haventory/lifecycle_probe.py
+++ b/.claude/skills/run-haventory/lifecycle_probe.py
@@ -203,10 +203,7 @@ async def check_resources() -> list[Result]:
             ("pinned", "/haventory_static/haventory-card.js?v=deadbeefcafe"),
             ("stale", "/haventory_static/haventory-card.js?v=0.0.0"),
             ("bare", "/haventory_static/haventory-card.js"),
-            # Installs predating the move of the bundle into the integration
-            # package: the entry must migrate in place, not gain a sibling.
-            ("legacy", "/local/haventory/haventory-card.js"),
-            ("legacy-versioned", "/local/haventory/haventory-card.js?v=0.0.1"),
+            ("query", "/haventory_static/haventory-card.js?v=1&foo=bar"),
         ]
         for label, url in variants:
             ws = await connect(session)
--- a/custom_components/haventory/__init__.py
+++ b/custom_components/haventory/__init__.py
@@ -97,12 +97,6 @@ _CARD_BUNDLE_PATH = _WWW_DIR / _CARD_FILENAME
 _STATIC_URL_PATH = "/haventory_static"
 _CARD_URL_PATH = f"{_STATIC_URL_PATH}/{_CARD_FILENAME}"
 
-# Installs predating the move loaded the card from a copy in the config `www/`
-# tree. That copy goes away with the integration, so such an entry is ours to
-# rewrite rather than somebody else's resource to leave alone.
-_LEGACY_CARD_URL_PATH = "/local/haventory/haventory-card.js"
-_CARD_URL_PATHS = frozenset({_CARD_URL_PATH, _LEGACY_CARD_URL_PATH})
-
 # hass.data[DOMAIN] keys that outlive a config entry: the static route cannot be
 # unregistered, and the module URL has to be removed as the exact string it was
 # registered under.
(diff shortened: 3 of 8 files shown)


chrreiter:

| step | result |
| --- | --- |

PR #630, opened 2026-08-24
fix(media): remove an item's attachment directory once its last file is gone

chrreiter:
## Summary

`<config>/haventory/attachments/<item_id>/` outlived the item it belonged to: the delete
paths unlinked files and stopped there, so every item that ever carried a photo or a manual
left one empty directory behind for good. #333 recorded that as a decision to make rather
than a defect; this takes its **option 2**.

- `_delete_blocking` now finishes by `rmdir`-ing the directories its unlinks left empty, and
  `_sweep_blocking` does the same for a directory the setup sweep has itself just emptied.
  Both go through one `_prune_emptied_blocking(root, directories)`, in the executor with
  every other filesystem call.
- The `rmdir` is the whole guard. A directory that still holds anything — a file the
  operator put there, a tile from an encoder generation this build cannot name — refuses to
  be removed, so nothing has to recognise that file first; that is exactly the case the old
  `_sweep_blocking` docstring protected by leaving every directory alone. Any other refusal
  (a file being served at that instant, a mount point) leaves the directory where it was,
  which is where it was before this PR.
- "Last file" counts the row tile beside a picture: `async_delete_attachments` already
  unlinks the original and the tile as a pair, so the directory goes when the pair does.
  A directory kept because an old-generation tile was still in it is emptied and removed by
  the next setup sweep — that tile is the sweep's orphan.
- The media root itself is never a candidate, and neither is a directory resolving outside
  it: only a directory under the root is pruned.

`docs/data_shapes.md`'s on-disk-layout bullet gains the one sentence a user reading their
config directory needs. No WebSocket schema, error code or field moves.

Closes #333

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that changes existing behavior)
- [ ] Documentation / tooling / CI
- [ ] Refactor (no functional change)

## How was this tested?

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` → **1343 passed, 27 skipped**;
  `uv run ruff check .` → All checks passed; `uv run ruff format --check .` → 189 files already
  formatted; `uv run mypy` → no issues in 26 source files.
- [x] Frontend (`cards/haventory-card`): `npx eslint .`, `npm run typecheck`, `npx vitest run`
  → **1917 passed (67 files)**, `npm run build` → built (card untouched by this PR; built so
  the phacc frontend cases run).
[...]

PR #631, opened 2026-08-24
test: one setup per phacc file, one helper per online smoke, and the phacc twins

chrreiter:
## Summary

Seven folds across the test tree, no behaviour change and no production line touched.

- **One `setup_entry` fixture** in `tests/integration/conftest.py` replaces the fourteen
  identical `_setup` helpers the phacc files each carried. It is a factory rather than an
  autouse fixture because several tests fill `hass_storage`, monkeypatch a module or open a
  bus listener before the entry may load. `test_services.py` read the repository off its own
  `_setup`; it asks `find_runtime` for it at the call site now.
- **`install_runtime(…, ws=True)` and `ws_hass()`** replace the six copies of "a stub Home
  Assistant, an entry with a runtime, the WebSocket commands registered" (five named
  `_make_hass`, one `_hass`). `test_ws_storage_error_offline.py` keeps a wrapper — its callers
  want the repository and the store too — and reads the store off the runtime instead of
  building a second one and swapping it in.
- **`tests/online_helpers.py`** holds the four helpers the three online smokes each carried
  byte-identical: the WebSocket URL from the base URL, connect-and-authenticate,
  read-until-this-id, and the message-id source.
- **`tests/date_helpers.day_offset`** replaces the three copies of `_utc_day_offset`. It reads
  `dt_util.now()` — the clock every answer it is checked against is measured from — instead of
  `datetime.now(UTC)`; the reminder spec's own `_today` now does too. That is the #586 shape:
  a test whose expected date comes from a second, different reading of the clock.
- **`tests/lovelace_helpers.py`** holds one storage-mode and one YAML-mode resource collection,
  where the registration tests and the removal tests each had their own.
- **`mountHost`** in the card's `test.utils.ts` replaces the `mountStore` → `mountComponent` →
  `{el, store, hass, sr}` sequence in the three store-backed host specs.
- **The ten offline twins** of `tests/integration/test_frontend.py` go, and with them the stub's
  log of every `async_register_panel` call, which real Home Assistant does not have.

Refs #230 (TESTS-T5, TESTS-T6)

## Type of change

## Counting

| | Lines |
| --- | ---: |
| Production removed | 0 |
| Test removed | 1,105 |
| Added | 809 |

`git diff --stat origin/main`: 38 files changed, 809 insertions(+), 1105 deletions(-). Nothing
under `custom_components/` or the card's non-test `src/` changed.

## What the twins were, and what is left

`tests/integration/test_frontend.py` runs on every pull request (`ci.yml`'s `integration` job)
and asserts these against a real Home Assistant, so the offline versions were a second, weaker
oracle for the same ten facts:

| Offline test dropped | phacc test that pins it |
| --- | --- |
| `test_serves_the_bundle_directory_without_cache_headers` | `test_bundle_is_served_without_cache_control` |
| `test_registers_both_loaders_with_the_same_url` | `test_both_loaders_get_the_same_url` |
| `test_static_path_is_registered_once_across_a_reload` | `test_reload_registers_the_static_route_once` |
| `test_unload_hands_back_the_module_url_and_setup_restores_it` | `test_unload_hands_back_the_module_url` |
| `test_registers_the_sidebar_panel_against_the_card_bundle` | `test_the_sidebar_panel_lands_in_the_real_panel_registry` |
| `test_a_reload_leaves_the_panel_in_place` | `test_a_reload_leaves_the_same_panel_object_in_place` |
| `test_a_disabled_entry_takes_the_sidebar_entry_back` | `test_a_disabled_entry_gives_the_sidebar_panel_back` |
| `test_removing_the_entry_takes_the_sidebar_entry_back` | `test_removing_the_entry_gives_the_sidebar_panel_back` |
| `test_a_rename_across_a_reload_replaces_the_panel` | `test_a_rename_replaces_the_panel_without_the_overwriting_error` |
| `test_an_explicit_opt_out_registers_no_panel` | `test_an_opted_out_entry_registers_no_panel` |

Kept, because nothing else can reach them: the cross-language pins (icon set, media route and
its parameters, status vocabularies, pills), the Lovelace resource migration, no bundle built,
no `http`, a refused route, and the panel toggle and rename through the options listener.

What a contributor loses is the ability to check those ten facts on a host that cannot run
phacc without Docker.

## How was this tested?

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` (1333 passed, 27 skipped),
  `uv run ruff check .`, `uv run ruff format --check .`, `uv run mypy` — all green.
  `npx vitest run` (1917 passed, 67 files), `npm run build` — all green.
- [x] phacc (`scripts/test_integration.sh`, card built first): **153 passed in 14.69s**.
- The online smokes cannot run here (no live Home Assistant, `RUN_ONLINE` unset). Their fold is
  proven by import and collection — `pytest --collect-only` over the three files collects 22
  tests — plus the offline gate, which imports every one of them.

## Checklist

- [x] Tests added/updated — this is a subtraction: tests are deleted or folded onto shared
  helpers, never rewritten to pass.
- [x] Docs updated where behavior changed — no behaviour changed; no file left
  `custom_components/haventory/`, so `RETIRED_PATHS` is unchanged, and no WS schema, error code
  or field moved.
  `docs/data_shapes.md` in sync — no API change.
  optimistic `version`).

- **The stub's `async_register_panel` stays; its `__panel_registrations__` log goes.** TESTS-T5
  lists the whole 60-line helper. Five panel tests with no phacc twin still need a
  `panel_custom` module — the options toggle, the live rename, a missing `panel_custom`, and no
  bundle built — and TESTS-T5 keeps them. What goes is the part real Home Assistant does not
  have: the log of every registration attempt. `test_applying_twice_over_registers_once` reads
  the registry instead, and the same `Panel` object after a second apply says the registration
  was neither replaced nor taken away and put back. `hass.http.static_path_calls` went the same
  way — its one reader was a dropped twin, and aiohttp's duplicate-route raise, which the stub
  keeps, is the real oracle.
- **Five `_make_hass` and one `_hass`, not six `_make_hass`.** M1's PR deleted
  `tests/test_ws_helpers_offline.py`, which held one of the six the audit counted.
  `tests/test_ws_reminders_offline.py`'s `_hass` is the same three lines under another name, so
  the count is six again and it is folded here.
- **`mountHost` serves three host specs, not four.** `haventory-panel.test.ts` is handed a
  `hass` and builds its own store; mounting it through `mountHost` would give it a second store
  nothing reads.
- **`test_ws_storage_error_offline.py` keeps a local `_make_hass`.** Its eight call sites want
  `(hass, repo, store)`; it is three lines over `ws_hass` now and is no longer a copy of
  anything.

## Follow-ups

- Sixty-five places in the offline suite still write `install_runtime(hass)` and
  `ws_setup(hass)` on two lines inside a test. The new `ws=` flag collapses each pair to one
  line, but those files are outside this package's list and the sweep is worth its own pass.
- `tests/integration/{test_areas,test_diagnostics,test_persistence,test_repairs}.py` build their
  entries inline with variations the `setup_entry` fixture does not cover (a setup expected to
  fail, custom `data`). Some of those inline blocks are the fixture verbatim and could move onto
  it.
- `tests/integration/test_sensor.py` has its own `_local_day_offset`, already reading
  `dt_util.now()`. It is the same four lines as `tests/date_helpers.day_offset`; a module both
  modes can import would be one more fold.

None of the three is filed: they are tidiness inside the test tree, not something a household
can meet.

## Handover

**1. Merged / left open** — this PR, for the M2 master to merge once its review is done.
Nothing here is left open for the owner.

**2. Test this by hand** — nothing. No user-visible surface, no stored data and no lifecycle
path changed; the ten dropped tests are covered by the phacc job CI runs on every pull request.

**3. Validate locally** — nothing. Every fold is pinned by a test that ran here: the offline
gate (1333 passed), the card gate (1917 passed) and `scripts/test_integration.sh` (153 passed).
The online smokes' fold is pinned by import and collection only, so the next session that has a
live Home Assistant can run `RUN_ONLINE=1 HA_BASE_URL=… HA_TOKEN=… uv run pytest -q -m online`
`[dev-ha]` and expect the same results as before this PR.

**4. Decisions taken against drifted notes** — see the section above: the stub keeps
`async_register_panel` and loses `__panel_registrations__`; five `_make_hass` plus one `_hass`
rather than six `_make_hass`; `mountHost` for three host specs, not four;
`test_ws_storage_error_offline.py` keeps a three-line local wrapper.

**5. Follow-ups** — the three named above, none filed.

**6. State left behind** — branch `claude/v0-8-0-m2-test-folds`; no assets branch; no Home
Assistant instance touched and no `.env` left in the worktree. Counting for this PR: production
removed 0, tests removed 1,105, added 809.

### DIFF
--- a/cards/haventory-card/src/components/hv-card-shell.test.ts
+++ b/cards/haventory-card/src/components/hv-card-shell.test.ts
@@ -5,7 +5,7 @@ import {
   makeItem,
   makeMockHass,
   mountComponent,
-  mountStore,
+  mountHost,
   q,
   settle,
   stubViewport,
@@ -32,15 +32,11 @@ function loc(id: string, name: string, parentId: string | null = null): Location
 }
 
 async function mountShell(opts: { items?: Item[]; locations?: Location[]; mobile?: boolean } = {}) {
-  const { hass, store } = await mountStore({
-    items: opts.items ?? [],
-    locations: opts.locations ?? [],
-  });
-  const { el, sr } = await mountComponent<HVCardShell>('hv-card-shell', {
-    store,
-    forceMobile: opts.mobile ?? false,
-  });
-  return { el, store, hass, sr };
+  return mountHost<HVCardShell>(
+    'hv-card-shell',
+    { items: opts.items ?? [], locations: opts.locations ?? [] },
+    { forceMobile: opts.mobile ?? false },
+  );
 }
 
 describe('hv-card-shell: header', () => {
--- a/cards/haventory-card/src/components/hv-full-view.test.ts
+++ b/cards/haventory-card/src/components/hv-full-view.test.ts
@@ -1,5 +1,5 @@
 import './hv-full-view';
-import { componentCss, makeItem, mountComponent, mountStore, q, settle, stubViewport } from '../test.utils';
+import { componentCss, makeItem, mountHost, q, settle, stubViewport } from '../test.utils';
 import { deepActiveElement, deepFocusables } from '../ui/dialog-focus';
 import { discardPrompt } from '../ui/discard';
 import { toIsoDate } from '../ui/relative-time';
@@ -32,16 +32,15 @@ async function mount(
     narrow?: boolean;
   } = {},
 ) {
-  const { hass, store } = await mountStore({
-    items: opts.items ?? [],
-    locations: opts.locations ?? [],
-    areas: opts.areas ?? [],
-    ...(opts.statuses ? { statuses: opts.statuses } : {}),
-  });
-  const { el, sr } = await mountComponent<HVFullView>(
+  return mountHost<HVFullView>(
     'hv-full-view',
     {
-      store,
+      items: opts.items ?? [],
+      locations: opts.locations ?? [],
+      areas: opts.areas ?? [],
+      ...(opts.statuses ? { statuses: opts.statuses } : {}),
+    },
+    {
       columns: ['quantity', 'category'],
       ...(opts.embedded ? { embedded: true } : {}),
       ...(opts.narrow ? { narrow: true } : {}),
@@ -49,7 +48,6 @@ async function mount(
     },
     { renders: 2 },
   );
-  return { el, store, hass, sr };
 }
 
 describe('hv-full-view: phone-width app bar', () => {
--- a/cards/haventory-card/src/components/hv-organize-dialog.test.ts
+++ b/cards/haventory-card/src/components/hv-organize-dialog.test.ts
@@ -1,6 +1,6 @@
 import { setLanguage } from '../i18n';
 import './hv-organize-dialog';
-import { all, componentCss, makeItem, mountComponent, mountStore, q, settle } from '../test.utils';
+import { all, componentCss, makeItem, mountHost, q, settle } from '../test.utils';
 // The clipboard itself is `ui/clipboard`'s own test; what the dialog owes is
 // asking the helper and believing its answer, which needs both answers.
 vi.mock('../ui/clipboard', async (importOriginal) => ({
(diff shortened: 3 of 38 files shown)


chrreiter:

| step | result |
| --- | --- |

PR #632, opened 2026-08-24
refactor: remove the WebSocket rate limiter

chrreiter:
## Summary

Deletes the WebSocket rate limiter, everything that configured it and everything that
reported on it. It shipped opt-in and off by default; the only callers that ever turned it
on were two harnesses in this repository, and it cost a token-bucket module, nine
config-entry options, a form section in three dictionaries, a branch in the command guard,
two branches in the event fan-out, an error code, a `haventory/health` sub-block, a
diagnostics block, a docs page, and — on the card — a retry loop, a banner, a bulk-failure
string and two drop-counter tiles.

What replaces it: nothing. The API is reachable only by someone already logged in to Home
Assistant, which is what the README's "Known limitations" now says in the entry the
rate-limiting one became.

Behaviour a household sees:

- The options form loses its (default-collapsed, default-off) rate-limit section. An entry
  that still carries the nine keys loads and ignores them, and the next save of that form
  drops them — `async_create_entry` replaces the stored options wholesale.
- `rate_limited` leaves the error taxonomy; `haventory/health` loses `rate_limit`; the
  diagnostics payload loses its `rate_limit` block and the runtime loses `rate_limiter`
  from `data_keys`.
- The card loses the "Rate limited" banner and its "Busy — retrying" sibling, whose counter
  only the command retry loop ever wrote. The **subscribe-retry** path stays whole — it
  serves `unavailable`, which is what a config-entry reload answers — and so does the
  `retryAfterHintMs` reader (M5 owns that one).
- A household that had the limiter on loses the limit.

`rate_limit.py` is appended to `RETIRED_PATHS`, and `docs/rate_limiting.md` is deleted with
its five inbound links.

Refs #230 (item 7), Refs #199

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [x] Breaking change (fix or feature that changes existing behavior)
- [ ] Documentation / tooling / CI
- [ ] Refactor (no functional change)

The breaking half is the feature's removal, not a wire-format change to anything that
stayed: no command, result field or event moves except the two the limiter owned.

## Counting

| | lines |
|---|---|
| production removed | 766 (backend 527 · card 239) |
| test lines removed | 1065 |
| lines added | 276 |

Docs −188 and skills −735 on top, from `git diff --stat`: 56 files, +276 / −2754.

## How was this tested?

[...]

Issue #542, opened 2026-08-21
Audit the 806 translation keys: what nothing reads, and what says the same thing twice


chrreiter:
### What & why

V0.7.0 shipped 806 translation keys — 719 in `cards/haventory-card/src/i18n/en.ts` and 87 in
`custom_components/haventory/translations/`. That is a high number for a card this size, and
it was never checked: the keys were extracted surface by surface, and nothing asked afterwards
whether a key was reached at all or whether the sentence next to it already existed.

**Do this before #540.** Every key kept is a key that has to be reworded, re-read and
re-reviewed there, and rewording a string that turns out to be dead — or one of five keys
saying "Tags" — is work spent twice. #540's tables are the review surface; this issue is what
decides how many rows they should have.

It also sits squarely in V0.8.0's optimization theme. The dictionaries are **71 516
characters** of keys and values, and the bundle grew from 637 kB to 708 kB across the three
i18n PRs (gzip 153 → 168 kB) — about 11%, carried by every dashboard on every load, in both
languages whether or not the reader can use either.

### What the measurement already says

Run against `main` at f1461df. Every number below is reproducible from `en.ts` / `de.ts` and a
sweep of the non-test sources.

**Two keys nothing reads.** Neither appears in any source file, by literal or by computed key:

| key | English |
| --- | --- |
| `hv.term.checkedOutUntil` | `Checked out · due {date}` |
| `hv.action.apply` | `Apply` |

Both were written for a call site that ended up composing the sentence differently. Nothing
catches this today — `catalog.test.ts` checks that every key is *translated*, never that any of
them is *used*.

**103 keys sit in 42 groups whose English and German are both identical.** Collapsing each
group to one key removes **61 keys, 8.5% of the card's set**. The largest groups:

| text | keys |
| --- | --- |
| Tags | `hv.editor.field.tags`, `hv.column.tags`, `hv.filter.tags`, `hv.fullView.tags`, `hv.organize.tab.tags` |
| Next inspection | `hv.sheet.fact.nextInspection`, `hv.editor.nextInspection`, `hv.column.inspection_date`, `hv.filter.sortField.inspection_date` |
| Reminder | `hv.sheet.fact.reminder`, `hv.editor.reminder`, `hv.column.reminder_date`, `hv.filter.sortField.reminder_date` |
| Updated | `hv.sheet.fact.updated`, `hv.column.updated_at`, `hv.filter.dateNoun.updated`, `hv.filter.sortField.updated_at` |
[...]

Issue #231, opened 2026-08-03
Frontend subtraction: comment archaeology, duplicated shell chrome, dead code


chrreiter:
### What &amp; why

The 2026-08-03 review's frontend verdict: well-engineered but ~1,500–2,000 LOC (~10–12%) removable with zero behavior change, and one class of comment that actively feeds the "AI-generated" impression. In priority order:

1. **Strip past-tense debugging-history comments** (~600–900 lines) — "The host measured scrollWidth 874…" (`hv-data-table.ts:96-111`), "It used to be an outlined 12.5px pill" (`hv-item-editor.ts:642-646`), the CSS forensics blocks in `hv-full-view.ts:283-382, 574-604` and `hv-card-shell.ts`. These violate CLAUDE.md's own comments policy (constraints, present tense, no history). **Recommended before the repo is read publicly** — the single highest-leverage de-slop edit.
2. **Shared shell chrome**: the four stat badges are written out twice (`hv-card-shell.ts:705-748` vs `hv-full-view.ts:1529-1572`), as are the debounced search, `_priceStaged`, and `_onEditorSave` (~150–250 LOC) — extract `renderStatBadges(counts, filters, variant)` + shared search/save helpers.
3. **Dead code**: the `pendingOps` chain (`store.ts:1000-1047`, `types.ts:480`, the never-bound `hv-list` `pendingIds`/selection props); `WSClient.ping/getLocation/updateCustomFields` (`ws.ts:38-40, 148-162, 185-187`); `@lit-labs/virtualizer` — a runtime dependency **never imported**. For the virtualizer: either adopt it for #191's virtual-scroll bullet (`loadAllPages(200)` can put 10k rows in the DOM) or drop the dependency; don't ship it unused.
4. **Store dedup**: a generic `optimisticWrite(fn, patch)` for the six structurally identical mutators (`store.ts:1051-1121`) + table-driven `refresh*`; single promise-unwrap helper in `ws.ts` (~90–130 LOC).
5. **Merge `store.test.ts` into `store.revamp.test.ts`** (the filename itself narrates history) and dedupe (~300–500 test LOC).

### Source

Release review 2026-08-03 (frontend review, simplification table); relates to #191 (virtual scroll).

### Impact

Medium (readability, contributor experience, "slop" optics)

### Effort

M (as a series of S PRs; item 1 is mechanical)

chrreiter:
[...]

PR #633, opened 2026-08-24
refactor(card): delete the dead selection chain, the unused client methods and the unbound props

chrreiter:
## Summary

The card's dead-code cut for V0.8.0 §6.M3, PR 1 — FE-STORE-S1/S2, FE-LISTS-L1/L8, FE-SHELLS-C7. Nothing here changes what a user sees; every target is either unreachable from any production path or a production API only a test called.

Refs #231

What goes:

- **`StoreState.pendingOps`** — seeded once, written in three `finally` blocks, read by nothing. An optimistic write shows its result on the row and rolls that row back when the call is refused, so no surface has a second per-operation state to draw. The `docs/frontend_architecture.md` state list drops it too.
- **Five `WSClient` methods** — `ping`, `addTags`, `removeTags`, `updateCustomFields`, `getLocation`. Tag and custom-field edits reach the backend through `items/bulk`, so the mock keeps its `item/add_tags`, `item/remove_tags` and `item/update_custom_fields` cases (the bulk handler dispatches them); only `location/get` goes. The backend commands themselves stay — `docs/backend_api_contract.md` describes a server surface, and the card not using one is not a reason to retire it.
- **The list's selection mode.** `hv-list`'s one host is `hv-card-shell`, which binds none of `selectable`, `selection`, `pendingIds`, `fill` or `skeletonRows`. So `hv-list-row`'s whole selection half is unreachable: the three props, the `.row.selected` rule, the 27 lines of `.box` CSS, the checkbox, the `toggle-select` emissions nothing listens for, the pending chip, and the gate that hid the hover actions behind it. Bulk selection lives on the full view's `hv-data-table`, which keeps its own checkbox column and its live `toggle-select` — untouched here. `fill` was never set, so the `:host(:not([fill]))` half is the branch that runs and the scroller's cap moves onto `.scroller`; `skeletonRows` becomes the constant it always was.
- **`hv-card-shell.forceMobile` and `ResponsiveController.setForced`** — a production property and a controller method that existed only because jsdom lays nothing out, so the shell's tests had no width to give it.
- **`MediaUrls.failed()`, `--hv-error-border`**, and four i18n names: `hv.row.select` and `hv.row.pending` (they die with the selection mode) plus `hv.term.checkedOutUntil` and `hv.action.apply` (#542's two already-unread names). All four go from both dictionaries and nothing else in them changes.

### The test-harness swap, in place of `forceMobile`

`stubElementWidth(width)` in `test.utils.ts` stands in a `ResizeObserver` that answers the moment a target is observed — which is inside `hostConnected`, so the first render already draws the mode the width implies. `hv-card-shell.test.ts` mounts at 375px or 1280px instead of pinning a boolean. **This is a harness swap, not a loosened assertion**: every assertion in that file is unchanged, and setting the phone constant to a desktop width fails 23 of its cases, so the width is really what decides the layout now.

`ui/media.test.ts` loses two `urls.failed(...)` assertions that restated the line above them. The third stood for "a refresh that failed is not remembered as *no URL is coming*", which the test now says in what the next read returns — a stronger observation than the accessor gave it.

### Two pins added

- `hv-list-row.test.ts` — a row opens the item it names and draws no `row-pending` chip beside the ones it does draw.
- `store.test.ts` — the store keeps no register of writes in flight.

Each carries a type line — an `Extract` of the removed name out of the type's keys, asserted to be `never` — as well as the rendered/runtime half, because a property no host binds and a state field nothing reads both cost zero failing tests when they come back. Both lines were checked in the failing direction: re-adding the field or the property makes `tsc --noEmit` fail.

### Counting

| | lines |
|---|---|
| Production removed | 210 |
| Test removed | 57 |
| Added | 155 |

`git diff --stat origin/main...HEAD`: 19 files, 155 insertions, 267 deletions. `hv-list-row.ts`'s 89 removed / 31 added is inflated by re-indenting the hover-actions block once its `selectable ? null :` wrapper went.

### Keeping #632 mergeable

PR #632 (the rate-limiter removal) is open on eight of the same files. Every target here sits outside its hunks, and nothing was reformatted, re-commented or re-imported near them.

```
$ git merge-tree --write-tree HEAD origin/claude/v0-8-0-m2-rate-limiter
3098496029b7a0ae8dbc633b209f5f60905a2756
$ echo $?
0
```

One tree OID, no conflict lines, exit 0 — against `a337dad`.

## Type of change

## How was this tested?

Both gates green before each of the six commits.

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` → 1333 passed, 27 skipped · `uv run ruff check .` → All checks passed · `uv run ruff format --check .` → 192 files already formatted · `uv run mypy` → no issues in 26 source files
- [x] Frontend (`cards/haventory-card`): `npx eslint .` clean · `npm run typecheck` clean · `npx vitest run` → 67 files, 1916 tests passed · `npm run build` → 727.76 kB

`scripts/test_integration.sh` not required: nothing under `custom_components/` or `tests/integration/` changes (the built bundle there is git-ignored).

## Checklist

- [x] PR title follows [Conventional Commits](https://www.conventionalcommits.org/) (e.g. `feat: …`, `fix: …`).
- [x] Docs updated where behavior changed (`README.md`, `docs/backend_api_contract.md`, `docs/data_shapes.md`).
- [x] WebSocket API changes keep `ws.py`, `docs/backend_api_contract.md`, and `docs/data_shapes.md` in sync.
- [x] Essential invariants preserved (case-insensitive search, denormalized `location_path`, optimistic `version`).

No WS schema, error code or field moves — the five client methods that go were calls the card never made, and the backend keeps every command. `docs/frontend_architecture.md` changes twice: the `StoreState` field list, and the testing note that named `setForced()`.

None — no visible change. The three states a screenshot would show (a hovered row's actions, the list's skeleton, the phone row) are covered by the card gate, and the master's live check after this PR is the browser confirmation.

1. **`hv-location-tree`'s `allLabel` / `allIcon` were not cut.** FE-L5 and the §6.M3 text call them unbound; they are bound today — `hv-organize-dialog.ts:1368` sets `allLabel`, `hv-item-editor.ts:1518-1519` sets both. Left alone.
2. **`toggle-select` is not dead everywhere.** `hv-full-view` binds `@toggle-select` on `hv-data-table`, whose `selectable` is driven by its own selecting state. Only `hv-list-row`'s emissions and `hv-list`'s forwarding go; `hv-data-table` is untouched.
3. **`MediaUrls.failed()` was still dead, and is cut.** The re-check asked for: the `failed` *field* on the cache entry is live, but it is `get()` that reads it — a signing that failed is answered from the entry instead of asked for again. The missing-picture state comes from `presence()`, not from `failed`. The *method* had zero callers outside `media.test.ts`, so the package's condition held. The class docstring, which claimed a caller drew a placeholder off it, is corrected rather than left wrong.
4. **The `healthCache` / `refreshHealth` chain is untouched**, as instructed — #632 still routes it to the diagnostics panel (`host-surfaces.ts:364`). Named under Follow-ups instead.
5. **`hv-list`'s `fill` and `skeletonRows` went with the three props #231 names** — same dead chain, per FE-L1, and the issue's estimate is short by that much.

## Follow-ups

- **`ui/media.ts`'s `Entry.failed`** is now read in exactly one place (`get()`). It is live and correct; a later pass could fold it into the `pending` check, but there is nothing to fix. Not filed.
- **`StoreState.selection`** sits beside the field this PR removed and is a set of ids the store seeds — unlike `pendingOps` it is genuinely read (the full view's bulk selection), so it stays. Recorded so the next reader does not re-derive it. Not filed.
- **The `healthCache` / `refreshHealth` chain** is close to dead on the card side once the diagnostics panel is the only consumer. Out of scope here and owned by #632's aftermath; the M1 backend health cut is what decides it. Not filed.
- **`en.ts`'s "Shared, added by the second component batch."** is comment archaeology sitting beside a line this PR edits. Left for §6.M3's PR 2, which is the comment sweep. Not filed.

## Handover

**1. Merged / left open** — this PR, for the M3 master to squash-merge once CI is green. Nothing is left open for the owner here; #632 stays the owner's, and this branch is proven mergeable under it.

**2. Test this by hand** — `[owner]` nothing specific. No text, layout or backend surface moves, and the German dictionary loses only keys English lost with it.

**3. Validate locally**
1. `[browser]` Open the card on a desktop dashboard and hover a row: the pencil and the ⋮ appear, the ⋮ opens, and its entries act as before. This is the one path the deleted `selectable ? null :` gate stood in front of.
2. `[browser]` Click a row: the item opens (inline editor on desktop, detail sheet on a phone). No checkbox appears anywhere on the card's list.
3. `[browser]` Load the card with a slow backend, or before the first list resolves: five skeleton rows, and the scroller still caps at its usual height rather than running to the page's bottom.
4. `[browser]` Open the item editor on the desktop card: the form gets the taller cap and its Save/Cancel row is reachable — the `:host([editing])` branch after the `[fill]` collapse.
5. `[browser]` At a phone width, the card renders its mobile layout — the measured-width path, which no longer has a forced mode behind it.

**4. Decisions taken against drifted notes** — the five above.

**5. Follow-ups** — the four above, none filed; none clears the real-world bar.

**6. State left behind** — branch `claude/v0-8-0-m3-dead-code`, six commits, no assets branch, no HA instance touched. Counting for this session: production removed 210 · test removed 57 · added 155.

### DIFF
--- a/cards/haventory-card/src/components/hv-card-shell.test.ts
+++ b/cards/haventory-card/src/components/hv-card-shell.test.ts
@@ -8,6 +8,7 @@ import {
   mountHost,
   q,
   settle,
+  stubElementWidth,
   stubViewport,
 } from '../test.utils';
 import { discardPrompt } from '../ui/discard';
@@ -31,13 +32,26 @@ function loc(id: string, name: string, parentId: string | null = null): Location
   };
 }
 
+// The shell measures its own element, so the two layouts are two widths. A
+// phone is 375px; 1280 is any dashboard column wider than the breakpoint.
+const PHONE_WIDTH = 375;
+const DESKTOP_WIDTH = 1280;
+
+let restoreWidth: (() => void) | null = null;
+
 async function mountShell(opts: { items?: Item[]; locations?: Location[]; mobile?: boolean } = {}) {
-  return mountHost<HVCardShell>(
-    'hv-card-shell',
-    { items: opts.items ?? [], locations: opts.locations ?? [] },
-    { forceMobile: opts.mobile ?? false },
-  );
+  restoreWidth?.();
+  restoreWidth = stubElementWidth(opts.mobile ? PHONE_WIDTH : DESKTOP_WIDTH);
+  return mountHost<HVCardShell>('hv-card-shell', {
+    items: opts.items ?? [],
+    locations: opts.locations ?? [],
+  });
 }
+
+afterEach(() => {
+  restoreWidth?.();
+  restoreWidth = null;
+});
 
 describe('hv-card-shell: header', () => {
   it('shows the configured heading and the live stat badges', async () => {
@@ -534,10 +548,8 @@ describe('hv-card-shell: list and footer', () => {
     // Deliberately not `mountStore`: the skeleton is what shows before the
     // first list resolves, so this store must not be initialised.
     const store = new Store(makeMockHass({ items: [] }), { retryBaseMs: 0 });
-    const { el } = await mountComponent<HVCardShell>('hv-card-shell', {
-      store,
-      forceMobile: false,
-    });
+    restoreWidth = stubElementWidth(DESKTOP_WIDTH);
+    const { el } = await mountComponent<HVCardShell>('hv-card-shell', { store });
 
     const list = q(el, 'hv-list')!;
     expect(list.shadowRoot?.querySelector('[data-testid="list-skeleton"]')).toBeTruthy();
--- a/cards/haventory-card/src/components/hv-card-shell.ts
+++ b/cards/haventory-card/src/components/hv-card-shell.ts
@@ -316,8 +316,6 @@ export class HVCardShell extends LitElement {
   /** Required. The shell subscribes to it itself — see `connectedCallback`. */
   @property({ attribute: false }) store!: Store;
   @property({ type: String }) heading = DEFAULT_CARD_TITLE;
-  /** Force a layout instead of measuring; `null` measures. */
-  @property({ attribute: false }) forceMobile: boolean | null = null;
   /**
    * Which quick-filter pills this dashboard offers, or `null` for all of them.
    * Passed on to the full view unchanged — one vocabulary on both surfaces.
@@ -431,7 +429,6 @@ export class HVCardShell extends LitElement {
       this._storeUnsub = this.store.state.onChange(() => this.requestUpdate());
       this._searchDraft = this.store.state.value.filters.q;
     }
-    if (changed.has('forceMobile')) this.responsive.setForced(this.forceMobile);
     // Reflect the mode so child selectors and :host([mobile]) rules apply.
     this.toggleAttribute('mobile', this.mobile);
     this._syncPinnedItem();
--- a/cards/haventory-card/src/components/hv-list-row.test.ts
+++ b/cards/haventory-card/src/components/hv-list-row.test.ts
@@ -593,21 +593,41 @@ describe('hv-list-row: the phone line’s pieces', () => {
   });
 });
 
-describe('hv-list-row: selection mode', () => {
-  it('swaps row navigation for a checkbox', async () => {
-    const el = await mount({ id: 'item-1' }, { selectable: true });
-    const seen = captured(el, ['open-item', 'toggle-select']);
+/*
+ * Bulk selection lives on the full view's `hv-data-table`, which draws its own
+ * checkbox column. The card's list has one job — open the item — so a row here
+ * navigates on click and draws nothing that would claim otherwise, whatever a
+ * host sets on it.
+ */
+describe('hv-list-row: one job per row', () => {
+  it('opens the item on click and offers the hover actions', async () => {
+    const el = await mount({ id: 'item-1' });
+    const seen = captured(el, ['open-item']);
 
-    expect(q(el, '[data-testid="row-select"]')).toBeTruthy();
-    expect(q(el, '[data-testid="row-edit"]')).toBe(null);
+    expect(q(el, '[data-testid="row-edit"]')).toBeTruthy();
+    expect(q(el, '[data-testid="row-select"]')).toBe(null);
 
     (q(el, '[data-testid="list-row"]') as HTMLElement).click();
-    expect(seen).toEqual(['toggle-select']);
+    expect(seen).toEqual(['open-item']);
   });
 
-  it('reflects the selected state on the checkbox', async () => {
-    const el = await mount({ id: 'item-1' }, { selectable: true, selected: true });
-    expect(q(el, '[data-testid="row-select"]')?.getAttribute('aria-checked')).toBe('true');
+  it('renders no pending chip beside the ones a row does draw', async () => {
+    // The store applies a write to the row and rolls it back on refusal, so
+    // there is no second "in flight" state for a chip to name. The type line
+    // is the other half: a property no host binds draws nothing, and a test
+    // that only looked at the output would not notice one coming back.
+    const el = await mount({
+      id: 'item-1',
+      checked_out: true,
+      quantity: 0,
+      low_stock_threshold: 5,
+    });
+
+    expect(q(el, '[data-testid="row-checked-out"]')).toBeTruthy();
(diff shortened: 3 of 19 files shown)


chrreiter:
**L1 (2026-08-29) — M3's "Validate locally" (the #236 union), walked on the dev HA** (`main` at `94025f4`, 558-item seeded household, HA 2026.7.4, Chromium 1920 px and iPhone 15 emulation, 0.7.1 control on :8124).

| step | result |
| --- | --- |
| 1 hover a row: the pencil and ⋮ appear, the ⋮ opens and its entries act | **PASS** — both hidden before hover, both visible after; the menu lists `edit, check-out, delete` and each acted in the later steps. |
| 2 click a row: inline editor on desktop, detail sheet on a phone; no checkbox | **PASS** — `inline-editor` on the wide card, `card-detail-sheet` at 375 px; zero checkboxes and zero `row-select` on the card's list. |
| 3 five skeleton rows before the first list resolves; the usual cap; the open editor's taller cap with Save reachable | **PASS** — with the `item/list` result held back 4 s: 5 `.skeleton-row`, `max-height: 420px`; the editing scroller reads `760px` and Save sits at y≈680 of 1000. |
| 4 phone width: the mobile layout, measured, no forced mode | **PASS** — `hv-card-shell[mobile]` set, no `forceMobile` property on the element. |
| 5 two tabs: an item created in one appears in the other with its count | **PASS** — created through tab A's form, tab B shows the row and "Showing 50 of 558" → "… of 559" without a reload. |
| 6 a forced `conflict` on the editor's Save: rollback, banner, view latest, re-apply | **PASS on the mechanism, with a finding** — with the item's events dropped, Save answers `conflict`: nothing written (server stays v2), the banner reads *Someone else changed this item.* with *View latest* / *Re-apply my change*; *View latest* brings the server's copy in (v2, q15). *Re-apply my change* resends the **whole form** on the latest version, so the other side's quantity (15 → 20) came back as the form's stale value — and the ordinary save without any conflict does the same: filed as **#659** (pre-existing on 0.7.1, not this PR's; a fix PR is in flight from this session). The row's ± sends no `expected_version` (a delta write cannot go stale), so the `routeWebSocket` recipe only bites through the editor — noted in the handover. |
| 7 diagnostics: status line, since tile, three facts, note; Refresh updates | **PASS** — "No issues · live", Since last refresh, Subscriptions / Data loaded / Integration version, the healthy note; Refresh re-reads the counts. |
| 8 rename a location: sidebar tree, counts and paths follow | **PASS** — `Vorratsschrank` → `Vorratsschrank L1`: the sidebar row, its `1 / 49` count under the search and the table's location cell all follow live; renamed back. |
| 9 two oversized photos from the editor: each refused by the picker, no connection-lost banner | **PASS** — two `huge-*.jpg` above `max_attachment_bytes` each listed as failed beside its name; `degraded-offline` count 0. |
| 10 stop HA under an open card (~6 s): one connection-lost banner; start → gone, live updates resume | **PASS** — exactly one `degraded-offline` banner after `docker stop`; after `docker start` the banner is gone and an out-of-band create repaints the row. |
| 11 #634 / #635 | nothing to click through — the frontend gate (2015 vitest across 79 files) and the build are green on the merged tree. |

The `card_views.test.mjs` / `probe` / `surfaces` self-tests: 53/53 from the main clone's skill dir. Console: the only lines were HA's own (`Failed to fetch dynamically imported module …/frontend_latest/…` while HA was stopped; the phone emulation's `/unknown/node_modules/@babel/runtime/…` fetches, identical on 0.7.1 and absent from the card bundle).

Evidence: [`m3/06-conflict-banner.png`](https://raw.githubusercontent.com/chrreiter/HAventory/claude-assets-v0-8-0-l1/l1/m3/06-conflict-banner.png), [`m3/10-connection-lost.png`](https://raw.githubusercontent.com/chrreiter/HAventory/claude-assets-v0-8-0-l1/l1/m3/10-connection-lost.png), [`m3/03-skeleton.png`](https://raw.githubusercontent.com/chrreiter/HAventory/claude-assets-v0-8-0-l1/l1/m3/03-skeleton.png). The full L1 record is on #236.

PR #636, opened 2026-08-24
refactor(store): one optimistic write, one read path, one subscription opener

chrreiter:
## Summary

The store said the same thing several times over. This PR says each of them once, and cuts
the one chain nothing was listening to.

- **The six item mutators** were eleven identical lines apiece — apply the guess, send, take
  the server's copy, roll the row back on refusal. `optimisticWrite(itemId, patch, call,
  details?)` is that shape once. `updateItem` folds in with the `{itemId, changes}` its
  conflict banner reads; `bumpReminder` folds in with no patch, which is both what "no
  optimistic update" means and what leaves it nothing to roll back. `createItem` (ensure
  presence, no optimistic apply) and `deleteItem` (remove, then restore) differ in what they
  do to the list and stay as they are.
- **The four attachment wrappers** each awaited, applied and returned the item. `applyResult`
  does it once.
- **The location and status mutators** each called, then awaited the refresh their change
  earns. `afterLocationChange` and `afterStatusChange` name which refresh that is.
- **One read path.** `listItems` carried its own copy of `run()`'s classification, and
  `countMatching`, `listAllMatching`, `refreshItem`, the location/status mutators and the
  import/export calls reached the socket with none — so a success on one of them left a
  "connection lost" banner standing, and a run of transport failures on them counted toward
  nothing. They all go through `run()` now and the comment above the fetchers states the rule
  instead of a claim the code did not keep. The `inflight` de-dupe on the list stays.
- **One subscription opener.** `subscribe()` and `subscribeAreaRegistry()` each implemented
  the function-or-promise unwrap with early cancel; `openSubscription(msg, cb, opts)` is that
  once, keeping the safer of the two variants for both — the `try/catch` around an unsubscribe
  nobody holds any more, which `subscribe` had and the registry watch did not.
- **One timer, one stale-answer guard.** Four debounce slots, three cancel helpers and the
  connection-lost grace period were the same six lines with a different handle; five counters
  were the same "claim a turn, drop the answer once the turn has passed". `Coalesced` and
  `Latest` are each of those once, so `dispose()` cancels a list rather than four hand-written
  `if (handle !== null)` blocks. `setDegraded` compares the keys the patch names instead of
  five fields by name.
- **The dead health chain** (its own commit). `WSClient.health()`, `Store.refreshHealth()`,
  `StoreState.healthCache`, `host-surfaces`' `.health=` binding and `hv-diagnostics-panel`'s
  `health` property were written and never read: the panel's status line derives from
  `connected`/`degraded`, its note renders unconditionally, and `this.health` appears in
  neither its render nor its copy-report. The chain goes, and with it the mock's
  `haventory/health` case, `__setHealth`, `HealthPatch` and `HealthResult`. **The backend
  command, its `backend_api_contract.md` entry and the panel's rendered output are unchanged**
  — the refresh button still re-reads counts through `refreshAll`.

### What deliberately did not move

The attachment family stays off `run()`, and now says why in the code: an upload's failures
are HTTP, carry no backend error code, and belong to the picker that raised them, so counting
them would read a rejected file as a lost connection.

Refs #231 (item 4 — FE-STORE-S3/S4/S5/S8; item 3's card-side health remainder). The issue is
closed by #635.

## Type of change

## Counting

| | lines |
|---|---:|
| production removed | 359 |
| test removed | 53 |
| added | 355 (339 production · 3 test · 13 docs) |

`git diff --stat` against `origin/main`: 10 files changed, 355 insertions(+), 422 deletions(-).

## How was this tested?

Both gates green before every commit; no new tests, because nothing new happens — the guards
that already pin this behaviour are the assertion:

- `store.test.ts` — each mutator's optimistic apply and reconcile, `updateItem`'s conflict
  carrying `{itemId, changes}` and its rollback, `deleteItem`'s rollback, `bumpReminder`'s
  no-guess shape and its refusal, the location mutators' refresh pairs.
- `store.subscriptions.test.ts` — the four topics' scope and re-open rounds, the registry
  watch's retry, catch-up and dispose, a promise-returning `subscribeMessage`.
- `store.degraded.test.ts` — the subscribe budget, the grace period, the connection-lost
  counting, and every "stops … once disposed".
- `store.facets.test.ts` / `store.filters.test.ts` / `store.events.test.ts` — the coalesced
  refreshes and the seq guards' stale-response drops.
- `ws.test.ts` — the subscribe unwrap in both of Home Assistant's answer shapes.
- `hv-diagnostics-panel.test.ts` — the panel's rendered output, its refresh event and its
  copy-report, unchanged; `haventory-panel.test.ts` — the panel opening from the ⋮ menu.

Two test edits are mechanical, not rewrites to pass:

- `store.degraded.test.ts` "queues one entry for an outage however many calls fail" called
  `refreshHealth` as the fourth failing call; it calls `refreshVersion` instead — the same
  kind of read, the same count, the same assertions.
- `store.degraded.test.ts` "reads the inventory once the backend starts answering" refuses
  "exactly the loads `init` starts in parallel", which is 7 now that health is not one of
  them. The comment there already names that number as the thing to keep in step.

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` (1304 passed, 27 skipped),
  `uv run ruff check .`, `uv run ruff format --check .`, `uv run mypy` — all green.
  (1903 passed, 71 files), `npm run build` — all green.

phacc (`scripts/test_integration.sh`) not required: nothing under `custom_components/` or
`tests/integration/` is touched. CI's `integration` job runs it anyway.

This is subscription-shaped, so the harder floor applies: `card-smoke.yml` dispatched against
this branch, or the `two_tab` recipe on the master's HA, before merge.

## Checklist

- [x] Tests added/updated — none added (no behaviour change); the health cut deletes the case
  that read `healthCache`, and two degraded cases are adjusted mechanically as described.
- [x] Docs updated where behavior changed — `docs/frontend_architecture.md`: `healthCache` and
  `__setHealth` off their lists, `init()`'s warm-up no longer names health, the `run()`
  sentence carries its one exception, the `WSClient` paragraph stops promising wrappers for
  commands nobody calls, and `openSubscription` is named.
  `docs/data_shapes.md` in sync — no WS schema, error code or field moves. The card stops
  *calling* `haventory/health`; the command and its contract entry are untouched.

None — no rendered output changes. The diagnostics panel renders byte-identically: the
property that goes was never read by its template.

## Follow-ups

- `store.test.ts` and its five siblings still define `fast` and `flush` per file. Named as a
  follow-up by #635 and deliberately left alone here. Not filed: it is a test-helper tidy, not
  something a household can hit.
- `Store.dispose()` still has no caller in the card (`store.subscriptions.test.ts` says so in
  its own comment). Wiring it up or dropping it is a decision, not a cut. Not filed: nothing
  leaks today, because the element that owns the store outlives it.

## Handover

1. **Merged / left open** — this PR, open for the master to review and merge; it is PR 4 of
   §6.M3 and the last one in that package.
2. **Test this by hand** — nothing this session can do that a person must. `[owner]` The card
   against a real store, if the master's HA cannot be stood up.
3. **Validate locally**
   1. `[browser]` Open the card in two tabs. Create an item in one — the other's list shows
      the row and its footer count moves, without a reload.
   2. `[browser]` Force a `conflict` on `haventory/item/update` (the `routeWebSocket` recipe)
      and edit a row: the row snaps back to its old values and the banner names the conflict,
      with the "view latest" and "re-apply" actions working.
   3. `[browser]` Open the diagnostics panel: the status line, the "since last refresh" tile,
      the three facts and the note read exactly as before, and Refresh still updates the
      counts.
   4. `[browser]` Rename a location: the sidebar tree, its counts and the rows' paths all
      follow, one refetch each.
   5. `[browser]` Upload an oversized photo twice from the item editor: each failure is shown
      by the picker on the file that failed, and no "connection lost" banner appears.
4. **Decisions taken against drifted notes**
   - FE-S4 says a rate-limited read "would start retrying"; #632 removed the limiter and the
     retry loop, so routing the reads through `run()` changes only the connection-lost
     counting and the error normalization, which is what the degraded cases pin.
   - FE-S4 says "route the other direct calls the same way where they should count toward
     `connectionLost`". The attachment family is where they should not: an upload's HTTP
     errors carry no code, so two rejected files in a row would have raised a false "could not
     reach Home Assistant". That exception is in the code and in the doc.
   - FE-S8 names three seq guards; the subscribe round and the area-registry generation are
     the same mechanism, so `Latest` serves all five and `dispose()` invalidates the two that
     outlive it.
   - FE-S5 says the `try/catch` difference between the two unwraps is harmless either way; the
     merged opener keeps the safer one, so the registry watch now swallows a throwing
     unsubscribe it would previously have let escape into a promise nobody awaits.
   - The measurement's FE-S3 line counts predate #632 and #633; the numbers in the counting
     table are from this diff.
5. **Follow-ups** — the two above, neither filed, with reasons.
6. **State left behind** — branch `claude/v0-8-0-m3-store-refactor`, no assets branch, no HA
   touched. Counting for this PR: production removed 359 · test removed 53 · added 355.

---
_Generated by [Claude Code](https://claude.ai/code/session_014oboNMQzmHqX7YoyPmw466)_

### DIFF
--- a/cards/haventory-card/src/components/hv-diagnostics-panel.test.ts
+++ b/cards/haventory-card/src/components/hv-diagnostics-panel.test.ts
@@ -1,6 +1,6 @@
 import './hv-diagnostics-panel';
 import type { HVDiagnosticsPanel } from './hv-diagnostics-panel';
-import type { DegradedState, HealthResult, StatsCounts } from '../store/types';
+import type { DegradedState, StatsCounts } from '../store/types';
 import { mountComponent, q, settle } from '../test.utils';
 // The clipboard itself is `ui/clipboard`'s own test; what this panel owes is
 // asking the helper and believing its answer, which needs both answers.
@@ -18,15 +18,6 @@ const counts: StatsCounts = {
   no_location_count: 3,
 };
 
-function health(patch: Partial<HealthResult> = {}): HealthResult {
-  return {
-    healthy: true,
-    issues: [],
-    counts,
-    ...patch,
-  };
-}
-
 const NO_DEGRADATION: DegradedState = {
   connectionLost: false,
   reloading: false,
@@ -38,7 +29,6 @@ const NO_DEGRADATION: DegradedState = {
 async function mount(props: Partial<HVDiagnosticsPanel> = {}) {
   const { el } = await mountComponent<HVDiagnosticsPanel>('hv-diagnostics-panel', {
     open: true,
-    health: health(),
     counts,
     version: { integration_version: '0.0.1', schema_version: 4 },
     degraded: { ...NO_DEGRADATION },
--- a/cards/haventory-card/src/components/hv-diagnostics-panel.ts
+++ b/cards/haventory-card/src/components/hv-diagnostics-panel.ts
@@ -10,7 +10,7 @@ import { relativeTime } from '../ui/relative-time';
 import { nextZBase } from '../utils/zindex';
 import { DialogFocus } from '../ui/dialog-focus';
 import { copyText } from '../ui/clipboard';
-import type { DegradedState, HealthResult, StatsCounts, VersionInfo } from '../store/types';
+import type { DegradedState, StatsCounts, VersionInfo } from '../store/types';
 
 /**
  * Diagnostics.
@@ -189,7 +189,6 @@ export class HVDiagnosticsPanel extends LitElement {
   @property({ type: Boolean, reflect: true }) open = false;
   /** Phone viewport: rise from the bottom edge instead of centring. */
   @property({ type: Boolean, reflect: true }) mobile = false;
-  @property({ attribute: false }) health: HealthResult | null = null;
   @property({ attribute: false }) counts: StatsCounts | null = null;
   @property({ attribute: false }) version: VersionInfo | null = null;
   @property({ attribute: false }) degraded: DegradedState | null = null;
--- a/cards/haventory-card/src/host-surfaces.ts
+++ b/cards/haventory-card/src/host-surfaces.ts
@@ -357,7 +357,6 @@ export class HostSurfaces {
         data-testid="host-diagnostics"
         ?open=${this.diagnosticsOpen}
         ?mobile=${mobile}
-        .health=${st?.healthCache ?? null}
         .counts=${st?.statsCounts ?? null}
         .version=${st?.versionInfo ?? null}
         .degraded=${st?.degraded ?? null}
--- a/cards/haventory-card/src/store/store.degraded.test.ts
+++ b/cards/haventory-card/src/store/store.degraded.test.ts
@@ -259,7 +259,7 @@ describe('subscribeRetryDelayMs', () => {
     await store.adjustQuantity('1', 1);
     await store.adjustQuantity('1', 1);
     await store.refreshStats().catch(() => undefined);
-    await store.refreshHealth().catch(() => undefined);
+    await store.refreshVersion().catch(() => undefined);
 
     expect(store.state.value.errorQueue.map((e) => e.code)).toEqual(['connection_lost']);
     expect(store.state.value.degraded.connectionLost).toBe(true);
@@ -454,7 +454,7 @@ describe('Store: a card built while the backend cannot answer', () => {
       // Exactly the loads `init` starts in parallel, so the refusal window
       // closes with it and the recovery that follows is answered normally. The
       // empty-list assertion below is what catches this count going stale.
-      hass.__failNext(8, REFUSED);
+      hass.__failNext(7, REFUSED);
       // The subscribe is refused the same way the commands were: a restarting
       // instance has no `haventory/subscribe` registered yet either.
       hass.__failSubscribeNext(4, REFUSED);
--- a/cards/haventory-card/src/store/store.test.ts
+++ b/cards/haventory-card/src/store/store.test.ts
@@ -296,18 +296,6 @@ describe('Store', () => {
     expect(typeof store.state.value.statsCounts?.items_total).toBe('number');
   });
 
-  it('populates healthCache on init and refreshes on demand', async () => {
-    const hass = makeMockHass({ items: [makeItem({ id: '1', name: 'A' })] });
-    const store = new Store(hass, fast);
-    await store.init();
-
-    expect(store.state.value.healthCache?.counts.items_total).toBe(1);
-
-    await store.createItem({ name: 'B' });
-    await store.refreshHealth();
-    expect(store.state.value.healthCache?.counts.items_total).toBe(2);
-  });
-
   it('deleteLocation removes an empty location and refreshes caches', async () => {
     const hass = makeMockHass({ items: [] });
     const store = new Store(hass, fast);
--- a/cards/haventory-card/src/store/store.ts
+++ b/cards/haventory-card/src/store/store.ts
@@ -323,6 +323,60 @@ export interface StoreOptions {
   retryBaseMs?: number;
 }
 
+/**
+ * A deferred call that can only ever have one fire outstanding.
+ *
+ * Every waiting the store does is this shape: a burst of events must ask once,
+ * a retry must be droppable, and a disposed store must fire none of them. The
+ * handle is held here rather than beside each caller so `dispose` cannot cancel
+ * three of four and leave the fourth running against a store nobody holds.
+ */
+class Coalesced {
+  private handle: ReturnType<typeof setTimeout> | null = null;
+
+  /** True while a fire is booked — a grace period that must not be restarted. */
+  get pending(): boolean {
+    return this.handle !== null;
+  }
+
+  /** Book `run`, dropping any fire already booked. */
+  schedule(delayMs: number, run: () => void): void {
+    this.cancel();
+    this.handle = setTimeout(() => {
+      this.handle = null;
+      run();
+    }, delayMs);
+  }
+
+  cancel(): void {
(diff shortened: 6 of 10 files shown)


chrreiter:
Master's evidence for the subscription-shaped floor, all on this branch's bundle (head `77838d4`) against this session's §3 Home Assistant (2026.6.0, wheel install):

- **`card-smoke.yml` dispatched against the branch**: [run 32749708663](https://github.com/chrreiter/HAventory/actions/runs/32749708663) — `smoke (stable)` and `smoke (beta)` both green.
- **Live-update browser smoke** (Chromium): PASS — create, rename and delete each reflected live over the WS subscription, zero page errors.
- **Two-tab fan-out**, `two_tab.mjs` on both paths: the service path and the `--ws` control each PASS all six oracles (both tabs receive `items/created` and `stats/counts`, both repaint with no interaction).
- **Forced conflict** (the `routeWebSocket` recipe on `haventory/item/adjust_quantity`): quantity 1 → optimistic 2 → rolled back to 1 when the forged `conflict` lands, one banner rendered, zero page errors — byte-identical to the same probe's run on `main` before this branch.
- **Diagnostics panel** on `/haventory`: "No issues · live", the "Since last refresh" tile, the three facts, the healthy note and the Refresh button all render as on `main` — the removed `health` property was never read by the template, as the PR says.

Review notes: the case ledger differs from `main` by exactly the one dropped healthCache pin (1904 → 1903, no renames); the `run()` line drawn (attachment family excluded) is confirmed — `errorCode()` maps a code-less HTTP error to the transport code, so counting uploads would false-trip the connection-lost banner. Merging.

---

chrreiter:
**L1 (2026-08-29) — M3's "Validate locally" (the #236 union), walked on the dev HA** (`main` at `94025f4`, 558-item seeded household, HA 2026.7.4, Chromium 1920 px and iPhone 15 emulation, 0.7.1 control on :8124).

| step | result |
| --- | --- |
| 1 hover a row: the pencil and ⋮ appear, the ⋮ opens and its entries act | **PASS** — both hidden before hover, both visible after; the menu lists `edit, check-out, delete` and each acted in the later steps. |
| 2 click a row: inline editor on desktop, detail sheet on a phone; no checkbox | **PASS** — `inline-editor` on the wide card, `card-detail-sheet` at 375 px; zero checkboxes and zero `row-select` on the card's list. |
| 3 five skeleton rows before the first list resolves; the usual cap; the open editor's taller cap with Save reachable | **PASS** — with the `item/list` result held back 4 s: 5 `.skeleton-row`, `max-height: 420px`; the editing scroller reads `760px` and Save sits at y≈680 of 1000. |
| 4 phone width: the mobile layout, measured, no forced mode | **PASS** — `hv-card-shell[mobile]` set, no `forceMobile` property on the element. |
| 5 two tabs: an item created in one appears in the other with its count | **PASS** — created through tab A's form, tab B shows the row and "Showing 50 of 558" → "… of 559" without a reload. |
| 6 a forced `conflict` on the editor's Save: rollback, banner, view latest, re-apply | **PASS on the mechanism, with a finding** — with the item's events dropped, Save answers `conflict`: nothing written (server stays v2), the banner reads *Someone else changed this item.* with *View latest* / *Re-apply my change*; *View latest* brings the server's copy in (v2, q15). *Re-apply my change* resends the **whole form** on the latest version, so the other side's quantity (15 → 20) came back as the form's stale value — and the ordinary save without any conflict does the same: filed as **#659** (pre-existing on 0.7.1, not this PR's; a fix PR is in flight from this session). The row's ± sends no `expected_version` (a delta write cannot go stale), so the `routeWebSocket` recipe only bites through the editor — noted in the handover. |
| 7 diagnostics: status line, since tile, three facts, note; Refresh updates | **PASS** — "No issues · live", Since last refresh, Subscriptions / Data loaded / Integration version, the healthy note; Refresh re-reads the counts. |
| 8 rename a location: sidebar tree, counts and paths follow | **PASS** — `Vorratsschrank` → `Vorratsschrank L1`: the sidebar row, its `1 / 49` count under the search and the table's location cell all follow live; renamed back. |
| 9 two oversized photos from the editor: each refused by the picker, no connection-lost banner | **PASS** — two `huge-*.jpg` above `max_attachment_bytes` each listed as failed beside its name; `degraded-offline` count 0. |
| 10 stop HA under an open card (~6 s): one connection-lost banner; start → gone, live updates resume | **PASS** — exactly one `degraded-offline` banner after `docker stop`; after `docker start` the banner is gone and an out-of-band create repaints the row. |
| 11 #634 / #635 | nothing to click through — the frontend gate (2015 vitest across 79 files) and the build are green on the merged tree. |

The `card_views.test.mjs` / `probe` / `surfaces` self-tests: 53/53 from the main clone's skill dir. Console: the only lines were HA's own (`Failed to fetch dynamically imported module …/frontend_latest/…` while HA was stopped; the phone emulation's `/unknown/node_modules/@babel/runtime/…` fetches, identical on 0.7.1 and absent from the card bundle).

Evidence: [`m3/06-conflict-banner.png`](https://raw.githubusercontent.com/chrreiter/HAventory/claude-assets-v0-8-0-l1/l1/m3/06-conflict-banner.png), [`m3/10-connection-lost.png`](https://raw.githubusercontent.com/chrreiter/HAventory/claude-assets-v0-8-0-l1/l1/m3/10-connection-lost.png), [`m3/03-skeleton.png`](https://raw.githubusercontent.com/chrreiter/HAventory/claude-assets-v0-8-0-l1/l1/m3/03-skeleton.png). The full L1 record is on #236.

PR #639, opened 2026-08-25
refactor(card): the filter chrome, once

chrreiter:
## Summary

PR 3 of §6.M4. The compact card and the expanded view ask the same questions of the same store, and each wrote the controls that ask them: two copies of the five stat pills, the search box, the applied-filter chips, the filter panel's four handlers, the staged head row and the commit row under it, and two spellings each of the search window and the staged-count window.

**First commit is test-only.** PR 2's review asked for two pins its merge landed without, and this PR carries them ahead of the chrome work:

- the card's read sheet closes when the store reports its item removed (`ItemWorkspace.syncPinnedItem`'s `wasRemoved` path, which the full view already had a pin for and the card had none);
- the staged-filter drop on a viewport flip, now routed through `ViewportNarrow`'s change callback — the controller case pins the callback (fires with the new answer on a flip, not on connect), and the expanded view's case pins what it is for: after a rotation the head row counts the applied filters again, not a staged set the controls no longer carry. `stubViewport` grows an `announce`, since a flip is the only way to reach a host that drops state when the answer changes.

Both new pins were checked against the production code removed: each fails without it.

**Then the cut.** `ui/stat-badges.ts` holds the pill arithmetic — which counts there are, which the config allows, which have anything to say, what pressing one does — and `ui/filter-chrome.ts` the rest, with the search-box and head-row CSS exported the way `bannerStack` is. Functions, not a base class. `_setFilters` was a private one-liner on both hosts and is now the store call it wrapped.

Per §5, what each surface *calls* its controls stays a parameter, because that is what the browser harnesses and each shadow root's own CSS locate: `badge-*` against `full-badge-*`, `search-input` against `full-search`, `sheet-*` against `full-panel-*`, `.sheet-head` against `full-panel-head`, and the chip classes each surface dresses a hue in.

Refs #231 (item 2, FE-SHELLS-C2 / C8, TESTS-T9)

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that changes existing behavior)
- [ ] Documentation / tooling / CI
- [x] Refactor (no functional change)

## Counting

| | lines |
|---|---|
| production removed | 417 |
| test removed | 104 |
| added | 1140 |

[...]

PR #640, opened 2026-08-25
refactor(card): one modal chrome

chrreiter:
## Summary

V0.8.0 §6.M4 PR 4, FE-DIALOGS-1 and FE-DIALOGS-4.

The backdrop, the centring layer and the panel box were written five times — with two
backdrop alphas — and beside them five stacking bases, five focus syncs, five Escape
bindings and two answers to "who closes this?". `ui/modal.ts` now holds all of it:
`modalChrome` and `modalSheet` for the CSS, and a `Modal` reactive controller for the
z-base, the focus in and back out, the Escape binding and the panel markup. Adopters: the
organize dialog, the import sheet, the confirm, the column picker and the diagnostics
panel. `ui/dialog-sheet.ts` folded in as `modalSheet` — separate from the chrome because
the organize dialog takes the chrome and keeps its full-bleed phone page.

One close convention with it: a dialog reports the dismissal and never writes its own
`open`. That is what `hv-bottom-sheet` already documents — the host binds the property
from its own state, so Lit never writes back a value it committed itself, and a host with
a question to ask first could only put the dialog up again behind the binding's back. All
five hosts already cleared their own state on `cancel`, so nothing moves on screen.

`hv-confirm` gains the return-focus it lacked (its own commit, its own tests): it kept
landing the caret on the accepting button and left it wherever it fell when the question
was answered. Through the shared chrome it now records where focus came from — across a
shadow root, which is where every form that asks a discard question lives — and hands it
back on both answers. That retires `ConfirmSpec.onCancel`, the second half of
`ConfirmDiscard`, and the item editor's manual refocus after a declined attachment
removal; the editor's remaining `_refocus` is the lightbox-outlived-its-photo rescue,
where there genuinely is nowhere to return to.

Inside the organize dialog: three head rows, six commit rows and two touch action sheets
become one method each, and the seven-entry disclosure-reveal registry becomes a `ref`
callback. The ref fires when an element is built and not on the re-renders that follow,
which is what "newly open" means — each disclosure is `keyed` on what it is about so a
second row rebuilds it rather than reusing the one already there. The work waits a
microtask: a ref fires while its subtree is still detached, where a scroll has nothing to
act on and `focus()` is a silent no-op, and that also puts it after the chrome has landed
focus on the panel. The dialog's own opening update still moves nothing, which is the
"a re-opened dialog moves nothing" rule and its test.

Refs #231

## Counting

| | lines |
|---|---:|
| production removed | 1041 |
| test lines removed | 63 |
| lines added | 1210 |

`git diff --stat` against `origin/main`; ignoring whitespace the production halves are
+596 / −645, the difference being re-indentation in `hv-organize-dialog.ts`. Net
production is −49 lines, well short of the survey's −350 estimate, and the reason is
worth recording: the CSS and the plumbing cut as predicted (five backdrop/wrap/panel
blocks and one `dialog-sheet.ts` down to one 40-line block; five z-bases, focus syncs,
Escape bindings and panel markups down to one controller; a 77-line registry down to a
ref), but a shared renderer's call sites cost roughly what the markup they replace did.
The two action sheets are the clearest case: one shape instead of two, for about 45 lines
more than they occupied. The package asks for that fold by name and the duplication it
removes is real, so it stands — but the line count is not the argument for it.

## Type of change

- [x] Refactor (no functional change) — three of the four commits
- [x] New feature (non-breaking change that adds functionality) — the confirm dialog's return-focus

## How was this tested?

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` (1304 passed, 27 skipped), `uv run ruff check .`, `uv run ruff format --check .`, `uv run mypy` — all clean
- [x] Frontend (`cards/haventory-card`): `npx eslint .`, `npm run typecheck`, `npx vitest run` (1929 passed), `npm run build`

phacc: not required — nothing under `custom_components/` or `tests/integration/` changes
(the card's build output there is git-ignored). CI's `integration` job ran it anyway:
`153 passed in 13.10s`.

New and moved tests:

- `src/ui/modal.test.ts` — the five dialogs draw the same three layers with a named,
  modal panel; every dismissal is reported and none writes its own `open`; the phone flag
  reflects so the sheet rules can match (moved from `ui/dialog-sheet.test.ts`, which goes);
  each opening stacks over the last and re-opening takes a fresh base; focus goes in on
  open and back on close.
- `hv-confirm.test.ts` — focus returns to the control that raised the question, including
  one inside another shadow root, on both a declined and an accepted answer.
- `hv-organize-dialog.test.ts` — the value editor reveals itself again when the same row
  switches from rename to merge, which is what its key encodes. The existing
  disclosure block passes unchanged, including "moves nothing when the dialog re-opens
  with a disclosure still expanded".
- The three self-close assertions in the organize dialog's tests move onto the `cancel`
  event, which is now how the dialog says it is done; the equivalents in the column
  picker, the import sheet and the confirm are deleted, replaced by the one sweep in
  `modal.test.ts`.

## Checklist

- [x] Docs updated where behavior changed — `docs/frontend_architecture.md` names
      `ui/modal.ts` in place of `dialog-sheet.ts` in both the module table and the phone
      presentation paragraph.
      `docs/data_shapes.md` in sync — none in this PR.
- [x] Essential invariants preserved — nothing here touches search, `location_path` or
      `version`.

None from this session: it does not drive a Home Assistant. The visible surface is every
dialog's chrome, which the master's live check covers at 1920 px and 375 px in both
themes.

## Follow-ups

- **The centring grid's auto track.** `hv-diagnostics-panel` keeps two `grid-template-*`
  lines with its own comment: a `minmax(0, 1fr)` track is what gives `.panel`'s
  `max-width: 100%` and `max-height: 100%` something to resolve against. The other four
  want the same — the organize dialog's `max-height: 100%` is inert without it, so a tall
  dialog in a short desktop window runs off the bottom — but making it shared would change
  layout where the old rule was broken, which a cut PR should not do unseen. Worth a look
  in the live check. Not filed.
- **Focus after an attachment is removed in the editor.** Confirming the removal takes the
  button that raised the question out of the document, so focus lands on `<body>` and
  Escape no longer reaches the form until something is clicked. Unchanged by this PR — it
  behaved the same before — and `ui/attachments.ts` in PR 5 (FE-DIALOGS-5) is the natural
  place for the rescue. Not filed.
- **`hv-organize-dialog.ts` still renders a bare `Create`** outside `t()` in the value
  creator's footer. Known #542 drift; dictionaries and their call sites are M5's, so it is
  left exactly as found. Not filed (already counted in the FE-DIALOGS survey).

## Handover

1. **Merged / left open** — this PR is left for the master to squash-merge once its review
   and live check pass; no session merges its own.
2. **Test this by hand** — `[phone]` the organize dialog's phone page: the ⋮ action sheets
   on a location row and on a tag row, and that a row's editor still scrolls itself into
   view with the keyboard on its first field. `[phone]` the four smaller dialogs still
   rise from the bottom edge rather than arriving centred.
3. **Validate locally** — see the numbered block below.
4. **Decisions taken against drifted notes** — the survey proposed `renderModal({z, ...})`
   as a free function; the z-base and the "where focus came from" are per-instance state
   that outlives a render, so a host would have had to keep both and hand them back every
   render. It is a `ReactiveController` with a `render` method instead, which is the same
   three responsibilities in one place. The survey also expected `hv-confirm` to take the
   generic panel focus; it keeps landing the caret on its accepting button (Enter
   completes, Escape aborts), which the chrome supports as an `initialFocus` option.
6. **State left behind** — branch `claude/v0-8-0-m4-modal-chrome`, no assets branch, no HA
   touched. Counting: production removed 1041 · test lines removed 63 · added 1210.

### Validate locally

1. `[browser]` At 1920 px, open each of the five dialogs in turn — ⋮ → Columns, ⋮ →
   Diagnostics, ⋮ → Import, Organize, and any delete confirmation. Each arrives centred
   over a dimmed page. Expected: the same backdrop tone under all five (they were 0.4 and
   0.35 before), the same corner radius and shadow.
2. `[browser]` With each open, press Escape. Expected: it closes, and focus is back on the
   control that opened it — the ⋮ entry, the Organize button, the row action.
3. `[browser]` Raise a delete confirmation from inside the organize dialog (Tags tab → a
   row's ✕). Expected: the confirmation sits over the dialog, Escape closes only the
   confirmation, and focus returns to the ✕ that raised it.
4. `[browser]` Edit an item, type in a field, press Escape. Expected: the discard question
   appears with the caret on Discard; press Escape again and the caret is back in the
   field you were typing in, with the typing intact.
5. `[browser]` At 375 px, open the column picker, the import sheet, the diagnostics panel
   and a confirmation. Expected: each rises from the bottom edge, full width, with the
   home-indicator gap under its buttons. Organize is still a full-bleed page with a back
   arrow, not a sheet.
6. `[browser]` In the organize dialog, Locations tab, with enough locations to scroll:
   open the ✎ editor on a row near the bottom. Expected: the editor scrolls into view and
   the caret lands in its name field. Then open the ✎ on a different row — it scrolls
   again. Type in the name field: the pane does not move.
7. `[browser]` Tags tab: press ✎ on a row, then ⧉ (merge) on the same row. Expected: the
   form swaps to the merge shape, scrolls into view again and the caret lands in the
   target field.
8. `[browser]` Open the organize dialog, start a new status (Statuses tab → New), close
   the dialog with Escape, and re-open it. Expected: the status editor is still expanded,
   nothing scrolls, and focus is on the dialog panel — not in the editor's field.
9. `[browser]` At 375 px, Locations tab: tap a row's ⋮. Expected: the four-button sheet
   opens under the row, scrolls into view, and the caret does not move into it. Tap Edit —
   the sheet closes and the editor takes its place, scrolled to and focused.
10. `[browser]` Run `card_views.test.mjs` and `visual_pass.mjs` light and dark. Expected:
    every `data-testid` still resolves; the only pixel difference is the backdrop tone on
    the four dialogs that were at 0.35.


chrreiter:
**Master's review and live check (M4).**

Diff read in full. The behaviour change is properly isolated (`feat` commit for the confirm's return-focus, with its own tests), the retirement chain is coherent (`ConfirmSpec.onCancel` → gone, `ConfirmDiscard` back to one argument, the editor's `_refocus` kept only for the lightbox rescue), and the four disclosure pins survive in the organize spec, including "moves nothing when the dialog re-opens with a disclosure still expanded" plus a new mode-switch case. The `Modal`-as-controller and diagnostics-keeps-its-grid decisions are right. Counting reconciles (`-w` shows the organize churn is re-indentation; net production −49).

Live check on this session's HA (branch bundle byte-verified), light and dark:

- **Pixels**: `visual_pass` 42/42 per theme. Against the current `main` captures, the four dialogs that were at 0.35 (diagnostics, import, columns — card and panel) differ by a uniform max-delta of 13, exactly the 0.35→0.4 backdrop step; `d-14-full-columns`' larger deltas are confined to x 1368–1418 — the scrimmed timestamp column. Nothing else moved. The phone column picker gains the sheet's bottom gap (67 px at delta 4), which is the intended `modalSheet` form.
- **The behaviour commit, live**: Escape on a dirty form puts the caret on the accepting button; Escape again declines and the caret lands back in the name field with the typing intact — through `DialogFocus` across the shadow roots, no `onCancel` left anywhere.
- **Organize**: a bottom-row location editor scrolls into view with the caret in the name field; the Tags remove confirmation opens over the dialog with the caret on its accepting button; the phone diagnostics dialog rises full-width from the bottom edge.

Two observations checked A/B against `main`'s bundle and found **identical there** — pre-existing, not this PR's, recorded for L1 rather than filed:

1. A dialog launched from the ⋮ menu cannot return focus to its opener on close — the menu entry is removed from the DOM when the menu closes, so focus falls to the page. Equal on both builds.
2. Escape on the Tags-tab delete confirmation closes the confirmation *and* the organize dialog (declining by clicking Cancel does not). Equal on both builds; no data is lost either way.

The ✕-close/reopen resetting disclosures is also equal on `main` — the vitest re-open pin covers the path that leaves one expanded.

Captures: [`claude-assets-v0-8-0-m4/pr640/`](https://github.com/chrreiter/HAventory/tree/claude-assets-v0-8-0-m4/pr640) — backdrop before/after, the organize reveal, the phone sheet.

---

chrreiter:
**L1 (2026-08-29) — this PR's "Validate locally" (#639's seven steps and #640's ten), walked on the dev HA** (`main` at `94025f4`, 558-item seeded household, HA 2026.7.4, Chromium at 1920 px and as an iPhone 15, 0.7.1 control on :8124).

| step | result |
| --- | --- |
| #639-1 card 1920, light and dark: badge words/order, quiet total, `badge-low` filters and rings | **PASS** — `558 items · 19 low · 7 overdue · 7 to inspect · 4 to do · 10 checked out`, identical words and order on 0.7.1; pressing low lists the 19 and the badge carries `on`. |
| #639-2 card 375: the badge row on its own line, no total | **PASS** — five badges under the title (y 149/199 against a title bottom of 127), wrapping, no `badge-total`. |
| #639-3 phone filter sheet: head row, staging, `Show N items`, Cancel, Clear all | **PASS** — "Filters · 0 active · Clear all"; staging a category leaves the 50 rows behind the sheet untouched and the button reads *Show 12 items*; Cancel drops the draft; Clear all resets the sheet only (*Show 558 items*, sheet still up). |
| #639-4 expanded view 1920: five pills, the checked-out one translucent, search floor | **PASS** — five pills in order; low/inspect/to-do amber `rgb(255,167,38)`, overdue red `rgb(219,68,55)`, checked-out `rgba(255,255,255,0.22)`; the search box's 260 px floor on its container. |
| #639-5 `/haventory` 375: head and commit rows | **PASS** — Clear all flush with the head row's right edge (373/373); Cancel and the count pill inside the commit row. |
| #639-6 every harness `data-testid` resolves | **PASS** — 53/53 harness self-tests; `visual_pass` 42/42 in light and dark. |
| #639-7 `visual_pass` against the pre-merge captures | **PASS on the card header, search row, filter sheet, app bar and phone filter panel** — compared by item against the 0.7.1 control's captures; the whole-surface pixel diff is dominated by the seeds' second-granularity `updated_at` ordering (see the L1 handover), not by chrome. |
| #640-1 the five dialogs' chrome | **PASS** — Columns, Diagnostics, Import, Organize and a delete confirmation: backdrop `rgba(0, 0, 0, 0.4)`, radius 14 px, shadow `rgba(0,0,0,0.28) 0 12px 40px` on all five. |
| #640-2 Escape closes each; focus returns to the opener | **PASS** — the ⋮ trigger for the three menu entries, the Organize button for Organize; a confirmation launched from a row's ⋮ has no opener left (M4's follow-up, unchanged). |
| #640-3 a confirmation from inside Organize (Tags → ✕) | **Sits over the dialog (z 10007 > 10005): PASS. Escape closes only the confirmation: FAIL — and so does Cancel by click.** Both close the Organize dialog with the confirmation and land focus on the Organize button; a Location row's confirmation behaves. Identical on the 0.7.1 control, so not this PR's — filed as **#661**. |
| #640-4 dirty editor: Escape → caret on Discard; Escape → back in the field | **PASS** — `confirm-accept`, then `editor-name` with the typing intact. |
| #640-5 375 px: column picker, import, diagnostics, confirmation as sheets; Organize a page | **PASS** — import, diagnostics and the confirmation rise full-width (393/393) to the bottom edge; Organize is a full-bleed page with its back arrow. The phone card's own menu carries no column picker (`select-items, organize, refresh, diagnostics, export-all, import`) — the phone row has no columns; the picker was measured on the expanded view at 1920 px. |
| #640-6 Locations ✎ near the bottom | **PASS** — the editor scrolls into view (y 707 of 1000) with the caret in its name; another ✎ scrolls again; typing leaves the pane at scrollTop 229. |
| #640-7 Tags: ✎ then ⧉ | **PASS** — the merge shape, in view, caret in `value-target`. |
| #640-8 Statuses → New, Escape, re-open | **PASS** — the editor still expanded, nothing scrolled, focus on the dialog panel. |
| #640-9 375 px Locations ⋮ | **PASS** — the four-button sheet under the row (the pane scrolls it into view), the caret not in it; Edit swaps it for the editor, scrolled to and focused on the name. |
| #640-10 `card_views.test.mjs` and `visual_pass` light/dark | **PASS** — as above. |

Evidence: [`m4b/640-1-organize.png`](https://raw.githubusercontent.com/chrreiter/HAventory/claude-assets-v0-8-0-l1/l1/m4b/640-1-organize.png), [`m4b/640-3-confirm-over-organize.png`](https://raw.githubusercontent.com/chrreiter/HAventory/claude-assets-v0-8-0-l1/l1/m4b/640-3-confirm-over-organize.png), [`m4b/640-5-phone-import.png`](https://raw.githubusercontent.com/chrreiter/HAventory/claude-assets-v0-8-0-l1/l1/m4b/640-5-phone-import.png), [`m4b/639-3-phone-filter-sheet-staged.png`](https://raw.githubusercontent.com/chrreiter/HAventory/claude-assets-v0-8-0-l1/l1/m4b/639-3-phone-filter-sheet-staged.png), [`m4b/639-1-badges-light-0.7.1.png`](https://raw.githubusercontent.com/chrreiter/HAventory/claude-assets-v0-8-0-l1/l1/m4b/639-1-badges-light-0.7.1.png). The full L1 record is on #236.

PR #641, opened 2026-08-25
refactor(card): the widgets written twice

chrreiter:
## Summary

PR 5 of §6.M4 of `dev/V0_8_0_implementation.md` — FE-DIALOGS-2, -3, -5 and -7a from the
2026-08-22 comment on #231, plus the fold #578's follow-up named.

Five widgets that were written twice now have one definition each:

- **`ui/day-offsets.ts`** (FE-DIALOGS-2) — the three presets, the "+X days" box and the rule
  that an empty box clears the date, drawn once for the check-out popover and the editor's
  inspection field. Test-id stem and touch sizing stay per surface.
- **`ui/location-picker.ts`** (FE-DIALOGS-3) — the trigger, the `aria-expanded` /
  `aria-controls` pair, the holder that outlives the tree in it, the open state and
  close-on-pick, for the editor's location field, the filter panel's Where chip and the
  organize dialog's parent picker and merge target. `keepOpenOnSelect` is the filter panel's
  "adding to a set means picking again" as a flag.
- **`ui/attachments.ts`** (FE-DIALOGS-5) — the photo figure, the document row and the
  lightbox host the editor and the detail sheet share, including the one answer both give to
  a reference whose file the backend does not have (#606).
- **The category list opens in flow** (FE-DIALOGS-7a) — the floating placement, its
  measurement, its two window listeners and its stacking base go; the chevron and the
  show-all it opens stay.
- **`ui/roving-list.ts` grows open/close and parent stepping** — `hv-location-tree`'s own
  `_walk` / `_syncRovingTabindex` / arrow switch fold onto it.

Two behaviour changes ride along, each with its own commit and its own test, per §5: the
category list moving in flow (the owner's §2 item 3 neighbour decision — "keep the control;
stop it floating") and the focus that used to strand on `<body>` after an attachment removal
is confirmed in the editor, which PR 4's report named as belonging here.

Refs #231 #578 #606

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue) — the stranded focus after a
      confirmed attachment removal
- [x] Documentation / tooling / CI — `docs/frontend_architecture.md`'s `ui/` map

## Counting

| | lines |
| --- | ---: |
| production removed | 659 |
| test lines removed | 12 |
| lines added | 1,502 (886 production, 616 test) |

From `git diff --numstat origin/main`. Net production is **+227**, and it is worth saying
where it went rather than rounding it off: the six components lose **638** lines and gain
**309** (net −329), and the four `ui/` modules cost **556**, about a third of that their own
doc blocks. The issue's −125 estimate for the picker assumed an element that also absorbed
`hv-location-tree`'s properties; each of the four pickers passes a different set of them, so
that half of the saving does not exist. The saving that does exist is one definition per
widget, and 128 lines of new module tests that pin the shared rules directly.

- **The picker is a controller, not an element** (FE-DLG-3 says "a small element"). Two
  reasons, both in the module's header. Its trigger wears `.field-button`, `hv-chip toggle`
  or `.control` and its holder `.tree-holder` at four different heights — rules that live in
  each host's own stylesheet and do not reach across a shadow boundary, so an element would
  have had to move that CSS or render in light DOM. And an element puts an update boundary
  between the host and the tree, so the tree arrives one microtask after the host's
  `updateComplete` — which is what every one of these surfaces' existing tests awaits; a cut
  whose tests need a second `await` is a cut that rewrote its tests. Written as a controller
  in the `Modal` style, the CSS and the timing are unchanged by construction. This also
  matches PR 3's "functions, not a base class".
- **`hv-bulk-bar` is not an adopter.** The plan lists it, and the #231 comment marks it
  "holder only": it has no trigger, no open state and no close-on-pick — an always-open tree
  in a `.tree-holder` div. Folding it would mean a picker mode with no trigger that is never
  closed, for three lines. Left alone.
- **`.tree-holder` CSS stays per host.** The markup moved; the rule did not, because the four
  holders differ (`max-height` 200/220/220/230, and two carry `margin-top`). Sharing it would
  have meant an override in three hosts and a split of the editor's `.tree-holder,
  .list-holder` rule.

## Behaviour that changed on purpose

1. **The category list opens in flow** under its input, the way the location tree in the same
   form already does, and pushes the form down instead of covering it. Its own commit, its
   own test; the old "floats the list" test is deleted because the behaviour it pinned is
   gone.
2. **Focus is caught after a confirmed attachment removal.** The guard hands focus back to
   the control that raised it — the removed tile's own remove button, still in the document
   when the guard closes and gone a moment later when the strip redraws. The form now puts
   focus back on the name field, the same rescue the lightbox already had. `focusStranded()`
   is the test `DialogFocus` was already making, now named and shared.

Two visual details the folds settle, neither of them a behaviour (the precedent is PR 4's
backdrop alpha): the check-out popover's offset chips gain the editor's `cursor: pointer`,
and the editor's document rows now carry the `missing` class the sheet's do — the editor has
no rule behind it and draws exactly as before.

## How was this tested?

Both gates green before every commit.

- [x] Backend: `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run pytest -q` (1304 passed, 27 skipped),
      `uv run ruff check .`, `uv run ruff format --check .`, `uv run mypy` (25 files, clean)
      (76 files, 1968 tests passed), `npm run build`

CI: all 12 checks green, `integration` included.

phacc locally: **not required** — nothing under `custom_components/` or `tests/integration/`
changes, and §6.M4 marks the package "No phacc." CI's `integration` job ran it anyway and is
green.

New and grown tests: `src/ui/day-offsets.test.ts` (8), `src/ui/location-picker.test.ts` (7),
`src/ui/attachments.test.ts` (12), `src/ui/roving-list.test.ts` (+9 for the disclosure and the
riders), `src/ui/dialog-focus.test.ts` (+2 for `focusStranded`), and one case each in the
editor (focus after removal), the filter panel and the organize dialog (where the holder
lands relative to the row its trigger sits in — the one regression this PR nearly shipped and
now pins). No test was rewritten to pass: the only one deleted is the editor's "floats the
list over the form", whose behaviour is gone.

#575's keyboard cases on both roving hosts and #578's one-tab-stop cases are untouched and
are the pin for the roving fold.

## Checklist

- [x] Docs updated where behavior changed — `docs/frontend_architecture.md`'s `ui/` map no
      longer says the tree carries its own copy of the roving pattern, and names the three new
      modules.
      `docs/data_shapes.md` in sync — no backend change in this PR.
- [x] Essential invariants preserved — card-only refactor; no store, payload or query path
      touched.

None from this session: it has no Home Assistant and never talks to one. The master's live
check and the "Validate locally" block below cover the visible surfaces.

## Follow-ups

- **An orphaned doc block in `hv-item-editor.ts`** — "Checking out asks for a due date;
  checking in just happens…" sits directly above `_renderInspectionOffsets` with no symbol of
  its own, describing the check-out popover instead. Pre-existing on `main`; noticed because
  this PR shortened the block underneath it. Not filed: a comment defect, no user-visible bar.
- **The organize dialog's merge row still carries two inline `style` attributes** — the
  source chip's `text-decoration: line-through` and the row's own
  `display:flex;align-items:center;gap:11px;flex-wrap:wrap`. This PR moved the third
  (`flex:1;min-width:180px` → `.control.grow`) because the picker needs the trigger to carry
  classes and nothing else. Not filed: cosmetic, clears no real-world bar.
- **`hv-checkout-popover` still hand-rolls its own `DialogFocus`, `nextZBase` and scrim**
  rather than taking `ui/modal.ts` — it is anchored to a control rather than centred, which
  is why PR 4 left it out. Not filed: the module would need an anchored mode, and that is a
  package of its own.

## Handover

1. **Merged / left open** — this PR is for the master to review and merge; nothing is left
   open on purpose.
2. **Test this by hand** — `[phone]` the editor's location picker, category list, day offsets
   and photo strip on `/haventory` and in the card: the category list now pushes the form
   down instead of floating over it, which is the one thing a 375 px screen shows differently
   from 0.7.1. `[phone]` the filter panel's Where chip: the tree still opens under the chip
   row, not beside the Area select.
3. **Validate locally** — the numbered list below.
4. **Decisions taken against drifted notes** — the three under "Decisions" above.
6. **State left behind** — branch `claude/v0-8-0-m4-shared-widgets`, no assets branch, no HA
   touched. Counting: production removed 659, tests removed 12, added 1,502.

### Validate locally

1. `[browser]` **The editor's category list, at 1920 px and at 375 px, light and dark.**
   Open an item's editor on the card, in the full view and on `/haventory`, focus the
   Category field or press its chevron. Expected: the list opens **under the input, in flow**,
   pushing the fields below it down — no floating layer, and scrolling the surface behind it
   never leaves the list behind. The chevron still shows every category, typing still narrows,
   arrows still walk it, Escape still shuts it before it shuts the form.
2. `[browser]` **The editor's location field on all three hosts.** Press it: the tree opens in
   the box under it, `aria-expanded` flips, picking a location closes it and writes the path
   onto the button. Escape closes the tree first and the form second. Creating the first
   location from an empty tree still files the item in it and closes the tree.
3. `[browser]` **The filter panel's Where chip, desktop panel and phone sheet.** The tree
   opens **under the chip row**, full width, never beside the Area select. Multi-selecting
   leaves it open; "All items" closes it.
4. `[browser]` **The organize dialog's parent picker and merge target.** Parent picker: the
   tree opens under the control inside the location editor; picking a location or an area
   closes it. Merge: the target control keeps the rest of the row beside the struck-through
   source, and its tree opens **under that row**, not inside it.
5. `[browser]` **The check-out popover and the editor's inspection offsets.** Both draw the
   same three chips and the same "+X days" box; pressing a preset marks it, "+X days" takes
   the mark and opens the box, emptying the box clears the date (and disables the popover's
   confirm). At 375 px both grow for a thumb.
6. `[browser]` **The photo strip and the document rows, editor and detail sheet.** Thumbnails
   open the lightbox; a document opens in a new tab. With the media directory moved away,
   every picture draws the dashed "File missing" tile and every document row the amber chip —
   on **both** surfaces, and no broken-image glyph anywhere.
7. `[browser]` **Remove a photo from the editor and press Escape straight after.** Expected:
   the guard closes, the tile goes, and Escape still reaches the form (it asks the discard
   question or closes it) rather than doing nothing — that is the focus rescue.
8. `[browser]` **Keyboard-only through the locations tree**, in the full view's sidebar and in
   the organize dialog: one Tab stop into the tree, ↓↑ walk it, → opens a closed row and steps
   into an open one, ← closes an open row and steps out of a closed one, Home/End reach the
   ends, and Tab from the active row steps through that row's rename/merge/delete before
   leaving the tree. The count from the search box to the first table row is still 31 or
   fewer.

### DIFF
--- a/cards/haventory-card/src/components/hv-checkout-popover.ts
+++ b/cards/haventory-card/src/components/hv-checkout-popover.ts
@@ -4,12 +4,8 @@ import { customElement, property, state } from 'lit/decorators.js';
 import { tokens, base } from '../ui/tokens';
 import { onEscape } from '../ui/keyboard';
 import { icon } from '../ui/icons';
-import {
-  DEFAULT_CUSTOM_DAYS,
-  quickDayOffsets,
-  addDays,
-  formatDate,
-} from '../ui/relative-time';
+import { DEFAULT_CUSTOM_DAYS, addDays, formatDate } from '../ui/relative-time';
+import { dayOffsets, renderDayOffsets } from '../ui/day-offsets';
 import { nextZBase } from '../utils/zindex';
 import { DialogFocus } from '../ui/dialog-focus';
 import type { Item } from '../store/types';
@@ -41,6 +37,7 @@ export class HVCheckoutPopover extends LitElement {
   static styles = [
     tokens,
     base,
+    dayOffsets,
     css`
       :host {
         display: block;
@@ -93,51 +90,15 @@ export class HVCheckoutPopover extends LitElement {
         display: grid;
         gap: 8px;
       }
-      .offsets {
-        display: flex;
-        gap: 7px;
-        flex-wrap: wrap;
-      }
-      .offset {
-        border: 1px solid var(--hv-divider);
-        background: none;
-        color: var(--hv-chip-text);
-        border-radius: var(--hv-radius-chip);
-        padding: 6px 13px;
-        font: 400 12.5px var(--hv-font);
-      }
+      /* The shape is ui/day-offsets; a thumb's worth of height on top of it is
+         this popover's, and the editor that draws the same chips grows them by
+         its own amount. */
       :host([touch]) .offset {
         min-height: 40px;
         padding: 0 15px;
         font-size: 13.5px;
       }
-      .offset.on {
-        background: var(--hv-primary-dark);
-        border-color: var(--hv-primary-dark);
-        color: #fff;
-        font-weight: 500;
-      }
-      .custom {
-        display: flex;
-        align-items: center;
-        gap: 8px;
-        padding: 8px 12px;
-        border: 1px solid var(--hv-divider);
-        border-radius: var(--hv-radius-input);
-        font-size: 13px;
-        color: var(--hv-text-secondary);
-      }
-      .custom input {
-        width: 72px;
-        box-sizing: border-box;
-        border: 1px solid var(--hv-input-border);
-        border-radius: var(--hv-radius-input);
-        background: var(--hv-surface);
-        color: var(--hv-text);
-        padding: 5px 8px;
-        font: 400 13.5px var(--hv-font);
-      }
-      :host([touch]) .custom input {
+      :host([touch]) .day-box input {
         min-height: 44px;
         width: 88px;
         font-size: var(--hv-input-font, 14.5px);
@@ -325,52 +286,26 @@ export class HVCheckoutPopover extends LitElement {
           <div class="sub">${t('hv.checkout.sub')}</div>
         </div>
         <div class="body">
-          <div class="offsets">
-            ${quickDayOffsets().map((offset) => {
-              const value = addDays(offset.days);
-              return html`<button
-                class="offset ${!this._customOpen && this._due === value ? 'on' : ''}"
-                data-testid="checkout-offset"
-                data-days=${offset.days}
-                @click=${() => {
-                  this._customOpen = false;
-                  this._due = value;
-                }}
-              >
-                ${offset.label}
-              </button>`;
-            })}
-            <button
-              class="offset ${this._customOpen ? 'on' : ''}"
-              data-testid="checkout-offset-custom"
-              @click=${() => {
+          ${renderDayOffsets(
+            { current: this._due, customOpen: this._customOpen, customDays: this._customDays },
+            {
+              prefix: 'checkout',
+              onPick: (date) => {
+                this._customOpen = false;
+                this._due = date;
+              },
+              onCustom: (date) => {
                 this._customOpen = true;
-                this._due = addDays(this._customDays);
-              }}
-            >
-              ${t('hv.editor.customDaysOffset')}
-            </button>
-          </div>
-          ${this._customOpen
-            ? html`<label class="custom" data-testid="checkout-custom">
-                <input
-                  type="number"
-                  min="1"
-                  max="3650"
-                  inputmode="numeric"
-                  aria-label=${t('hv.editor.daysFromToday')}
-                  .value=${String(this._customDays)}
-                  @input=${(e: Event) => {
-                    const days = Number((e.target as HTMLInputElement).value);
-                    // An empty or nonsense box means no date yet, not a stale
-                    // one — the confirm button disables itself on a null due.
-                    this._customDays = days;
-                    this._due = Number.isFinite(days) && days >= 1 ? addDays(Math.floor(days)) : null;
-                  }}
-                />
-                <span>${t('hv.editor.daysFromToday')}</span>
-              </label>`
-            : null}
+                this._due = date;
+              },
+              // A cleared box leaves no due date, and the confirm button
+              // disables itself on one.
+              onDays: (days, date) => {
+                this._customDays = days;
+                this._due = date;
+              },
+            },
+          )}
           <label class="date ${this._due ? '' : 'none'}" data-testid="checkout-date">
             ${icon('calendar', 17)}
             <span class="hv-sr-only">${t('hv.editor.dueDate')}</span>
--- a/cards/haventory-card/src/components/hv-detail-sheet.ts
+++ b/cards/haventory-card/src/components/hv-detail-sheet.ts
@@ -22,13 +22,13 @@ import {
   pictures,
 } from '../ui/media';
 import type { MediaBindings } from '../ui/media';
+import { renderDocumentRow, renderLightboxHost, renderPhotoFigure } from '../ui/attachments';
 import type { ConfirmDiscard } from '../ui/discard';
 import { COPIED_MS, copyText } from '../ui/clipboard';
 import type { AreaRef, Item, Location, LocationTreeNode, MediaConfig, ScalarValue, StatusDefinition } from '../store/types';
 import './hv-bottom-sheet';
 import './hv-checkout-popover';
 import './hv-item-editor';
-import './hv-lightbox';
 import type { HVBottomSheet } from './hv-bottom-sheet';
 import type { HVItemEditor } from './hv-item-editor';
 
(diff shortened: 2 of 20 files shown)


chrreiter:
**Master's review and live check (M4).**

Diff read in full. The two behaviour commits are properly isolated with their own tests: the category move inverts the floating pin into an in-flow pin (no `floating` class, no inline placement, no `position: fixed` left in the CSS), and the focus rescue extracts `focusStranded()` in `ui/dialog-focus.ts` as report-not-act, with the editor deciding where focus belongs. The picker-as-controller decision is right for the stated reasons (host-stylesheet class vocabularies don't cross a shadow boundary; no second `await` forced into four suites).

The two open questions, answered: **`hv-bulk-bar` stays out** — FE-DLG-3 itself called it "holder only", and a no-trigger, never-closing picker mode for three lines is the wrong trade. **The module headers stay** — constraint comments are this repository's convention, not slack. The net production growth (+227, with the six components 329 lighter and the four `ui/` modules carrying the cost) is accepted and recorded, same category as #639's.

Live check on this session's HA (branch bundle byte-verified), 13/13 scripted checks plus the sweep:

- The category list opens in flow under its input, the form grows, scrolling never strands it, Escape closes the list before the form.
- The location picker on all four adopters: the editor's field (aria-expanded, pick writes the path and closes), the filter panel's Where chip (tree stays open across two picks), the organize parent picker (closes on a pickable row; the excluded self stays disabled). Merge target rides the same controller.
- The day-offset chips are identical on the editor and the check-out popover (`+7/+30/+90/+X days`), with the settled pointer cursor.
- With a probe item carrying real uploads and one file moved away on disk: the editor and the phone read sheet both draw the missing-file placeholder with no broken-image glyph anywhere; restored afterwards.
- The focus rescue live: removing a photo and confirming leaves focus on the editor's name field, not `<body>`.
- Keyboard: search box to the first table row in 28 presses (ceiling 31); the sidebar tree keeps one tab stop with ↓↑/Home/End walking it through the shared roving list.
- `visual_pass` 42/42 per theme; a same-store pixel compare against `main`'s bundle (the probe data had moved under the older captures) shows only the usual relative-timestamp regions on 8 surfaces — no chrome moved.

The attachment probes' 5 image-pipeline failures are environmental — the uploads re-encode correctly server-side (8.2 MB → 0.91 MB); the probe's byte-level read-back needs the Docker container this cloud session doesn't have.

Captures: [`claude-assets-v0-8-0-m4/pr641/`](https://github.com/chrreiter/HAventory/tree/claude-assets-v0-8-0-m4/pr641).

---

chrreiter:
**L1 (2026-08-29) — this PR's "Validate locally" (#641's eight steps and #644's six), walked on the dev HA** (`main` at `94025f4`, 558-item seeded household, HA 2026.7.4, Chromium at 1920 px and as an iPhone 15, 0.7.1 control on :8124).

| step | result |
| --- | --- |
| #641-1 the category list in flow, card and `/haventory` (the expanded view shares the editor) | **PASS** — `position: static`, 27 options; the Tags row below moves down 234 px on the card and 234 px on the panel while the list is open (on 0.7.1 nothing below moved — the list floated); Escape hides the list and leaves the form. |
| #641-2 the location field | **PASS** — the tree opens under the field with `aria-expanded="true"`, a pick closes it and writes the path onto the button; Escape closes the tree first, the form second. |
| #641-3 the filter panel's Where chip | **PASS** — the tree opens under the chip row (y 199 under a chip bottom of 186) at 1328 of the panel's 1360 px; two picks leave it open; *All items* closes it. |
| #641-4 organize parent picker and merge target | **PASS** — the parent tree opens under the control (868 under 853) and a pick closes it; the merge target's tree opens under the row (998 under a row bottom of 612 — the rest of the row stays beside the struck-through source). |
| #641-5 check-out popover and the editor's inspection offsets | **PASS** — three chips (+7 / +30 / +90 days) and the "+X days" box on both; a preset marks (`on`) and dates; the custom box takes the mark; emptying it reads *No due date* and disables Confirm. |
| #641-6 photo strip and document rows | **PASS** — a thumbnail opens the lightbox, a document opens with `target="_blank"`; with `attachments/` moved away every picture draws the dashed *File missing* tile and every document row the amber chip on the card row, the editor, the detail sheet and the table, and no `` renders broken. |
| #641-7 remove a photo, Escape straight after | **PASS** — the guard, the tile goes, Escape reaches the form (card and panel). |
| #641-8 keyboard-only through the tree; search → first row | **PASS on the sidebar** — one Tab stop (`tree-all` … `tree-orphans`), ↓↑→← Home/End move between rows and area heads, 31 presses from the search box to the first table row. **FAIL in the Organize dialog's Locations tab** — Tab from the active row goes to its count button and then the *next row's* count button; `tree-edit` / `tree-merge` / `tree-delete` keep `tabindex="-1"` while the row holds the stop. Identical on 0.7.1, so not this PR's fold — filed as **#663**. |
| #644-1 a dressed card row at 1920 (picture, low, flagged status, overdue loan) | **PASS** — `{thumb: 34px, Low, "Needs repair", "Overdue · Aug 10", height 58}`, byte-identical to the same item's row on the 0.7.1 control. |
| #644-2 the same item in the table | **PASS** — 34 px tile in the name cell, no Low chip, the status in its own column only, the checked-out chip reading just *Overdue*. |
| #644-3 `/haventory` at 375: the phone row | **PASS** — one line, 95 px with the stepper, elided path; the area pill is absent on the Kaffeefilter row by the rule that hides it when the area's name is the root location's (Küche / Küche), which #614 settled. |
| #644-4 keys on a card row; Delete inside the open ⋮ (#642) | **PASS** — Enter opens, Delete asks, `+`/`-` move 1→2→1; two Tabs reach the ⋮, Enter opens its three entries, Delete raises nothing behind it. |
| #644-5 keys on a table row; Enter on Edit | **PASS** — the same four keys, and Enter on the table's Edit opens the editor. |
| #644-6 a missing picture file on both surfaces | **PASS** — the dashed tile with the camera glyph and the *File missing* title on the card row and in the table. |

Evidence: [`m4c/644-1-card-row-main.png`](https://raw.githubusercontent.com/chrreiter/HAventory/claude-assets-v0-8-0-l1/l1/m4c/644-1-card-row-main.png) beside [`m4c/644-1-card-row-0.7.1.png`](https://raw.githubusercontent.com/chrreiter/HAventory/claude-assets-v0-8-0-l1/l1/m4c/644-1-card-row-0.7.1.png), [`m4c/641-1-card-category-in-flow.png`](https://raw.githubusercontent.com/chrreiter/HAventory/claude-assets-v0-8-0-l1/l1/m4c/641-1-card-category-in-flow.png), [`m4c/641-6-editor-missing.png`](https://raw.githubusercontent.com/chrreiter/HAventory/claude-assets-v0-8-0-l1/l1/m4c/641-6-editor-missing.png), [`m4c/641-3-filter-where-tree.png`](https://raw.githubusercontent.com/chrreiter/HAventory/claude-assets-v0-8-0-l1/l1/m4c/641-3-filter-where-tree.png). The full L1 record is on #236.

